PR #1, opened 2020-10-23
setup app.py

apuntamb:
This is the first edition of new code in this repo. This PR sets up the minimum of `.travis.yml`, a brand new `Pipfile` to start with and running the `app.py`.
In order to test the app.py changes, run:
 ```python
python3 app.py """To start the server"""
curl http://127.0.0.1:5000/ping """To call /ping endpoint"""
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: python
+
+python:
+  - "3.8"
+
+script:
+  - flake8
+
+install:
+  - pip install pipenv
+  - pipenv install --dev
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,14 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+flake8 = "*"
+
+[packages]
+Flask = "~=1.1.2"
+gunicorn = "~=20.0.4"
+
+[requires]
+python_version = "3.8"
--- a/app.py
+++ b/app.py
@@ -1,0 +1,12 @@
+from flask import Flask
+
+app = Flask(__name__)
+
+
+@app.route("/ping")
+def ping():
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: .travis.yml [hunk 1]
```diff
@@ -0,0 +1,11 @@
+language: python
+
+python:
+  - "3.8"
```
beav, line 4: hooray 3.8 :) 3.9 is the latest but is not supported everywhere quite yet.

Pipfile
```diff
@@ -0,0 +1,14 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+flake8 = "*"
+
+[packages]
+Flask = "*"
+gunicorn = "*"
```
patilsuraj767, line 11: Should we add restrict dependency? It will help to avoid issues caused due to breaking changes in dependent packages.

```
[packages]
Flask = "~=1.1.2"
gunicorn = "~=20.0.4"
```

apuntamb: makes sense, @beav thoughts?

beav: I've seen different projects do it both ways. As long as you remember to update to the latest versions once in awhile (perhaps make a monthly reminder in your calendar), it's OK to pin the versions.

kgaikwad: > Should we add strict dependency? It will help to avoid issues caused due to breaking changes in dependent packages.
> ```
> [packages]
> Flask = "~=1.1.2"
> gunicorn = "~=20.0.4"
> ```

I think it's good idea to lock these dependencies. As @patilsuraj767 mentioned let's point it to latest.

apuntamb: done.

patilsuraj767:
Tested pull-request locally. Works as expected :+1: 
Thank you @apuntamb

PR #2, opened 2020-11-09
Initial dev-setup for ros

patilsuraj767:
This will help in setting up the development environment for ROS.
With this PR I have also moved app.py inside ros/app.py (same like yupana and drift) to create room for other scripts and will separate out the core application.


.gitignore
```diff
@@ -103,6 +103,7 @@ celerybeat.pid
 
 # Environments
 .env
+!scripts/.env
```
kgaikwad, line 106: IIUC, this `.env` file contains configurations related to ingress service and not specific to ROS service so would it be good to rename it accordingly?

kgaikwad (changes requested):
@patilsuraj767,

Thank you, added one inline comment. 
Secondly, could you also verify attached sample file's content? I don't think it is good idea to include nested directories (Example - sample/{machine-fqdn}-20201108200821/var/tmp/metrics.json). Please correct me if I am missing anything.

patilsuraj767:
> @patilsuraj767,
> 
> Thank you, added one inline comment.
> Secondly, could you also verify attached sample file's content? I don't think it is good idea to include nested directories (Example - sample/{machine-fqdn}-20201108200821/var/tmp/metrics.json). Please correct me if I am missing anything.

Honestly, I have not thought regarding the tar extraction process. I have bundle it through the same way how normal insights-client and compliance does. They too have {machine-fqdn} in there tar archive. 

```
[root@client qh5i99eq]# insights-client --no-upload
Starting to collect Insights data for client.example.com
Archive saved at /var/tmp/gy8dq6r1/insights-client.example.com-20201202231217.tar.gz
[root@client qh5i99eq]# cd /var/tmp/gy8dq6r1/
[root@client gy8dq6r1]# ls
insights-client.example.com-20201202231217.tar.gz
[root@client gy8dq6r1]# tar xf insights-client.example.com-20201202231217.tar.gz 
[root@client gy8dq6r1]# ls insights-client.example.com-20201202231217/
blacklist_report  boot  branch_info  collection_stats  egg_release  etc  insights_commands  proc  run  sys  usr  var  version_info

```

kgaikwad:
> > @patilsuraj767,
> > Thank you, added one inline comment.
> > Secondly, could you also verify attached sample file's content? I don't think it is good idea to include nested directories (Example - sample/{machine-fqdn}-20201108200821/var/tmp/metrics.json). Please correct me if I am missing anything.
> 
> Honestly, I have not thought regarding the tar extraction process. I have bundle it through the same way how normal insights-client and compliance does. They too have {machine-fqdn} in there tar archive.
> 
> ```
> [root@client qh5i99eq]# insights-client --no-upload
> Starting to collect Insights data for client.example.com
> Archive saved at /var/tmp/gy8dq6r1/insights-client.example.com-20201202231217.tar.gz
> [root@client qh5i99eq]# cd /var/tmp/gy8dq6r1/
> [root@client gy8dq6r1]# ls
> insights-client.example.com-20201202231217.tar.gz
> [root@client gy8dq6r1]# tar xf insights-client.example.com-20201202231217.tar.gz 
> [root@client gy8dq6r1]# ls insights-client.example.com-20201202231217/
> blacklist_report  boot  branch_info  collection_stats  egg_release  etc  insights_commands  proc  run  sys  usr  var  version_info
> ```

Thank you @patilsuraj767 for this information.

kgaikwad (approved):

PR #5, opened 2020-12-16
Report consumer for processing ROS insights archive

patilsuraj767:


Makefile
```diff
@@ -4,8 +4,14 @@ help:
...
+		-H "x-rh-identity: eyJpZGVudGl0eSI6IHsiYWNjb3VudF9udW1iZXIiOiAiMDAwMDAwMSIsICJ0eXBlIjogIlVzZXIiLCAiaW50ZXJuYWwiOiB7Im9yZ19pZCI6ICIwMDAwMDEifX19Cg==" \
+		-H "x-rh-request_id: testtesttest" \
+		localhost:8080/api/ingress/v1/upload
+
 insights-upload-data:
-	curl -vvvv -F "upload=@sample.tar.gz;type=application/vnd.redhat.resource-optimization.sample+tgz" \
-		-H "x-rh-identity: eyJpZGVudGl0eSI6IHsiYWNjb3VudF9udW1iZXIiOiAiMTIzNDUiLCAiaW50ZXJuYWwiOiB7Im9yZ19pZCI6ICI1NDMyMSJ9fX0=" \
+	curl -vvvv -F "upload=@sample-files/insights-aws-sample.tar.gz;type=application/vnd.redhat.advisor.collection+tgz" -F 'metadata={"branch_info": {"remote_branch": -1, "remote_leaf": -1}, "bios_uuid": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71", "ip_addresses": ["10.74.255.52", "2620:52:0:4af8:21a:4aff:fe00:a8a"], "fqdn": "vm255-52.gsslab.pnq2.redhat.com", "mac_addresses": ["00:1a:4a:00:0a:8a", "00:00:00:00:00:00"], "satellite_id": -1, "subscription_manager_id": "7846d4fa-6fcc-4b84-aa13-5f12e588ecca", "insights_id": "eb6ed2ad-b395-4254-a94c-981d1b4957a4", "machine_id": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71"}' \
```
kgaikwad, line 14: `insights_id` key value is different than the value used for `insights_id` in command `ros-upload-data`. Is it correct?

kgaikwad, line 14: Do we really need to pass `metadata` here?

Makefile
```diff
@@ -4,8 +4,14 @@ help:
 	@echo "--- Commands for local development ---"
 	@echo "insights-upload-data       upload data to local ingress service for ROS processing"
 
+ros-upload-data:
+	curl -vvvv -F "upload=@sample-files/ros-sample.tar.gz;type=application/vnd.redhat.resource-optimization.sample+tgz" -F 'metadata={"branch_info": {"remote_branch": -1, "remote_leaf": -1}, "bios_uuid": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71", "ip_addresses": ["10.74.255.52", "2620:52:0:4af8:21a:4aff:fe00:a8a"], "fqdn": "vm255-52.gsslab.pnq2.redhat.com", "mac_addresses": ["00:1a:4a:00:0a:8a", "00:00:00:00:00:00"], "satellite_id": -1, "subscription_manager_id": "7846d4fa-6fcc-4b84-aa13-5f12e588ecca", "insights_id": "5ae6c008-b6b3-459b-82a0-7ebf75a047fd", "machine_id": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71"}' \
```
kgaikwad, line 8: IIUC, temporarily you have passed `metadata` details while request as tar is not having those details. 
will it be good to extract this metadata value in separate variable or json file? I found bit difficulty in reading this request. 
what do you think?

ros/lib/host_inventory.py
```diff
@@ -0,0 +1,10 @@
+import requests
+import json
+
+
+def fetch_host_from_inventory(insights_id, rh_identity):
+    host_api_url = f"http://localhost:8081/api/inventory/v1/hosts?insights_id={insights_id}"
```
kgaikwad, line 6: Worth extracting an inventory URL in a environment variable as URL will change as per environment (dev/qa/prod)?

ros/processors/report_processor.py
```diff
@@ -0,0 +1,110 @@
...
+from confluent_kafka import Consumer
+from ros.lib.host_inventory import fetch_host_from_inventory
+from ros.app import app, db
+from ros.models import PerformanceProfile
+
+LOG = logging.getLogger(__name__)
+running = True
+
+consumer = Consumer({
+    'bootstrap.servers': 'localhost:29092',
```
kgaikwad, line 16: Same here. It would be good to extract URL in config settings.

kgaikwad (changes requested):
Few inline comments.

Makefile
patilsuraj767: No, we do not need it in insights-upload-data command. I will remove it.

patilsuraj767: No, will fix it.

ros/processors/report_processor.py
```diff
@@ -0,0 +1,112 @@
...
+        consumer.close()
+
+    def handle_message(self):
+        self.report_url = self.msg.get('url', None)
+        metadata = self.msg.get('metadata', None)
+        if not metadata:
+            return None
+        insights_id = metadata['insights_id']
+        if not insights_id:
+            return None
```
kgaikwad, line 53:
```suggestion
        if not metadata or not metadata.get('insights_id'):
            return None
```
There is no any log statement or difference in return values so will it be good to handle both conditions in one if statement?

ros/processors/report_processor.py
```diff
@@ -0,0 +1,112 @@
...
+        rh_identity = self.msg.get('b64_identity', None)
+        host = fetch_host_from_inventory(insights_id, rh_identity)
+        if not host['results']:
+            print("No record found. Make sure system is registered in insights")
```
kgaikwad, line 57: Do we really need to mention message `Make sure system is registered in insights` here?

ros/processors/report_processor.py
```diff
@@ -0,0 +1,112 @@
...
+            return None
+        host_id = host['results'][0]['id']
```
kgaikwad, line 59: could you move this statement later after `report_url` checking?

ros/processors/report_processor.py
```diff
@@ -0,0 +1,112 @@
...
+        return performance_score
+
+    def _download_report(self):
+        download_response = requests.get(self.report_url)
+        if download_response.status_code != HTTPStatus.OK:
+            print("Unable to download the report")
+        return download_response.content
+
+    def _extract_performance_record(self, report_tar_gz):
+        tar = tarfile.open(fileobj=BytesIO(report_tar_gz), mode='r:gz')
```
kgaikwad, line 92:
```suggestion
        tar = tarfile.open(fileobj=BytesIO(report_tar_gz), mode='r:*')
```

On `yupana` service, recently we added support for `xz & bz2` along with `gz`, I think it would be good to have same support in ROS too. What do you think?

Accordingly, you will need to change variable naming.

kgaikwad (changes requested):

ros/lib/host_inventory.py
```diff
@@ -0,0 +1,11 @@
+import requests
+import json
+from ros.config import INVENTORY_ADDRESS
+
+
+def fetch_host_from_inventory(insights_id, rh_identity):
+    host_api_url = f"{INVENTORY_ADDRESS}/api/inventory/v1/hosts?insights_id={insights_id}"
+    headers = {'x-rh-identity': rh_identity, 'Content-Type': 'application/json'}
+    res = requests.get(host_api_url, headers=headers)
+    hosts = json.loads(res.text)
```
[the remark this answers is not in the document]
patilsuraj767: Yes, Changing it.

ros/processors/report_processor.py
```diff
@@ -0,0 +1,112 @@
...
+                consumer.commit()
+
+        consumer.close()
+
+    def handle_message(self):
+        self.report_url = self.msg.get('url', None)
+        metadata = self.msg.get('metadata', None)
+        if not metadata:
+            return None
+        insights_id = metadata['insights_id']
```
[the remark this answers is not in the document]
patilsuraj767: Exception block [here](https://github.com/RedHatInsights/ros-backend/pull/5/files#diff-3e77a430e2ccfd114f1e233d804c9077c12c508197b9ad61ab3038781f799b7aR40) will take care of it.

ros/processors/report_processor.py
patilsuraj767: :+1:

ros/processors/report_processor.py
patilsuraj767: Only printing `No record found` does not make much sense. With the message `Make sure system is registered in insights` user can proceed further and can try to register the system to insights.

ros/processors/report_processor.py
patilsuraj767: :+1:

ros/processors/report_processor.py
patilsuraj767: :+1:

ros/processors/report_processor.py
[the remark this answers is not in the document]
kgaikwad: What is the expected behaviour?

ros/processors/report_processor.py
```diff
@@ -0,0 +1,110 @@
...
+                else:
+                    record = PerformanceProfile(inventory_id=host_id,
+                                                performance_record=performance_record,
+                                                performance_score=performance_score)
+                    db.session.add(record)
+
+                db.session.commit()
+
+    def _calculate_performance_score(self, performance_record):
+        memory_score = (float(performance_record['avg_memory_used']) / float(performance_record['avg_memory'])) * 100
```
apuntamb, line 79: what if in any scenario if the data for a host is somehow not captured: 1. Is that a possible scenario? 2. If yes, shouldn't there be an exception handler here? If no, then we are good.

ros/processors/report_processor.py
```diff
@@ -0,0 +1,110 @@
...
+                if profile:
+                    profile.performance_record = performance_record
+                    profile.performance_score = performance_score
+                else:
+                    record = PerformanceProfile(inventory_id=host_id,
+                                                performance_record=performance_record,
+                                                performance_score=performance_score)
+                    db.session.add(record)
+
+                db.session.commit()
```
apuntamb, line 76: Should we catch the sqlalchemy errors, in case of any db exceptions?

apuntamb (changes requested):
Hey @patilsuraj767! Looks like a good one! :+1: 
Just added few inline comments/queries.

Makefile
```diff
@@ -4,8 +4,16 @@ help:
 	@echo "--- Commands for local development ---"
 	@echo "insights-upload-data       upload data to local ingress service for ROS processing"
 
+metadata={"branch_info": {"remote_branch": -1, "remote_leaf": -1}, "bios_uuid": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71", "ip_addresses": ["10.74.255.52", "2620:52:0:4af8:21a:4aff:fe00:a8a"], "fqdn": "vm255-52.gsslab.pnq2.redhat.com", "mac_addresses": ["00:1a:4a:00:0a:8a", "00:00:00:00:00:00"], "satellite_id": -1, "subscription_manager_id": "7846d4fa-6fcc-4b84-aa13-5f12e588ecca", "insights_id": "5ae6c008-b6b3-459b-82a0-7ebf75a047fd", "machine_id": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71"}
```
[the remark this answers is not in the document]
patilsuraj767: It is passed with the curl request, there is an `insights_id` field in metadata which helps backend to detect from which client system the request is coming from.

ros/config.py
```diff
@@ -1,12 +1,17 @@
...
-        self.db_user = os.getenv("ROS_DB_USER", "postgres")
-        self.db_password = os.getenv("ROS_DB_PASS", "postgres")
-        self.db_host = os.getenv("ROS_DB_HOST", "localhost")
-        self.db_port = os.getenv("ROS_DB_PORT", "15432")
-        self.db_name = os.getenv("ROS_DB_NAME", "postgres")
-        self.db_uri = f"postgresql://{self.db_user}:{self.db_password}"\
-                      f"@{self.db_host}:{self.db_port}/{self.db_name}"
+INVENTORY_HOST = os.getenv("INVENTORY_HOST", "localhost")
+INVENTORY_PORT = os.getenv("INVENTORY_PORT", "8081")
+INVENTORY_ADDRESS = f"http://{INVENTORY_HOST}:{INVENTORY_PORT}"
```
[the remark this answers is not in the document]
patilsuraj767: Yes, Sure

ros/processors/report_processor.py
patilsuraj767: I am not sure further going how insights-client will be collecting data and sending it to us. But if although there is any exception then this exception block [here](https://github.com/RedHatInsights/ros-backend/pull/5/files#diff-3e77a430e2ccfd114f1e233d804c9077c12c508197b9ad61ab3038781f799b7aR39) will handle it.

ros/processors/report_processor.py
patilsuraj767: If the only aim is to handle the exception and printing the log then again main exception block [here](https://github.com/RedHatInsights/ros-backend/pull/5/files#diff-3e77a430e2ccfd114f1e233d804c9077c12c508197b9ad61ab3038781f799b7aR39) in the parent method will handle it. 
Let me know if you are thinking of any counter step to perform in case of sqlalchemy exception then we can add the separate exception handling for it.

ros/lib/host_inventory.py
```diff
@@ -0,0 +1,10 @@
+import requests
+from ros.config import INVENTORY_URL
+
+
+def fetch_host_from_inventory(insights_id, rh_identity):
```
apuntamb, line 5:
```suggestion
def fetch_hosts_from_inventory(insights_id, rh_identity):
```

ros/lib/host_inventory.py
```diff
@@ -0,0 +1,10 @@
...
+    host_api_url = f"{INVENTORY_URL}/api/inventory/v1/hosts?insights_id={insights_id}"
+    headers = {'x-rh-identity': rh_identity, 'Content-Type': 'application/json'}
+    res = requests.get(host_api_url, headers=headers)
+    hosts = res.json()
+    return hosts
```
apuntamb, line 10: Also, can we rename this file to `host_inventory_interface.py` something like that? Just to get a better understanding  of what this file might contain.

apuntamb (changes requested):
Added few minor comments.

Makefile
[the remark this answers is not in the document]
apuntamb: Okk! So, based on this metadata information, does it create a separate metadata.json file?

ros/lib/host_inventory.py
patilsuraj767: In HBI every host coming through insights-client has the unique `insights_id` therefore, although this API call returns the list it will always have a single item. Do you think we should still rename it to  `fetch_hosts_from_inventory`?

apuntamb: Ok so, insights_id would be unique too? Then keeping it the same makes sense. `fetch_host_from_inventory`

Makefile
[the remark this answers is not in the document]
patilsuraj767: No, I don't think so. Basically, whatever we send in metadata to ingress, ingress puts that in the topic on ROS as a JSON object.

Pipfile
```diff
@@ -15,6 +15,7 @@ psycopg2-binary = "~=2.8.6"
 flask-migrate = "~=2.5.3"
 flask-script = "~=2.0.6"
 flask-cors = "~=3.0.9"
+confluent-kafka = "~=1.5.0"
```
apuntamb, line 18: There is `requests` module required too in the pipfile. Could you add it here?

kgaikwad: yeah :+1: 
Facing issue while running `python manage.py db upgrade ` - ModuleNotFoundError: No module named 'requests'

kgaikwad (commented):
Tested changes on local by making few manual steps. 
Out of which few needs to be fixed in PR:
- added `requests` in pipfile 
- currently inventory container is not running along with all other containers using `pipenv run docker-compose --env-file .env-minio up`.  
-  needs to run inventory API server separately

After this I am able to run both make commands and getting data populated in respective tables.

@patilsuraj767 could you do `requests` change in Pipfile? and can we modify docker-compose file to run inventory container as well in first go?

It would be great if we use different DB name other than postgres. This change won't block PR but just suggestion.

ros/processors/report_processor.py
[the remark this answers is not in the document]
patilsuraj767: It is expected that backend should ignore/reject reports with no insights_id.

apuntamb:
+1 for:
- modifying docker-compose.yml to update the imgress image tag and get the inventory container running,
- renaming db from postgres to something like `ros` or `ros-db`

Rest, PR looks almost ready now! :+1:

scripts/README.md
```diff
@@ -20,7 +20,13 @@ mkdir -p /usr/share/minio/{minio-conf,minio-data}
...
 # docker login quay.io
 # . .env-minio
-# docker-compose up
+# docker-compose --env-file .env-minio  up
+```
+
+4) Wait for insights-inventory container to get started and after that follow the below steps to start inventory API server.
+```
+# docker exec -it <insights-inventory-container-id> bash
+# python run_gunicorn.py
```
apuntamb, line 29: Should we also add a one-liner for `ros-upload-data`?

ros/processors/report_processor.py
```diff
@@ -0,0 +1,110 @@
...
+            performance_record = json.loads(record_string)
+            return performance_record
+
+
+def check_kafka_connection():
+    topics = consumer.list_topics().topics
+    if "platform.upload.resource-optimization" in topics:
+        return True
+    else:
+        return False
```
kgaikwad, line 110:
```suggestion
    return ("platform.upload.resource-optimization" in topics)
```
Using `in`, it returns True if element exists in list and False if the element does not exists in list.

ros/processors/report_processor.py
```diff
@@ -0,0 +1,110 @@
...
+import tarfile
+from io import BytesIO
+from http import HTTPStatus
+from confluent_kafka import Consumer
+from ros.lib.host_inventory_interface import fetch_host_from_inventory
+from ros.app import app, db
+from ros.config import INSIGHTS_KAFKA_ADDRESS
+from ros.models import PerformanceProfile
+
+LOG = logging.getLogger(__name__)
```
kgaikwad, line 13: I don't see anywhere in the processor file you ave used it.

kgaikwad (approved):
Added two inline nitpicks. 
@apuntamb, after these nitpicks,  :+1:  from my side for merge.

ros/processors/report_processor.py
patilsuraj767: Oh yes, I was going to implement the logger in ROS instead of using the print statement but I think we should now implement logger in some other PR. For now, I will remove the logger.

kgaikwad: Sounds good. Thank you @patilsuraj767

apuntamb (approved):

apuntamb:
Merged!
Thanks @patilsuraj767 @kgaikwad :+1:

PR #6, opened 2020-12-17
modify dummy data with key change & extra records

kgaikwad:


ros/resources/hosts.py
```diff
@@ -14,18 +14,18 @@ def get(self):
                     'id': '12345-57575757',
                     'account': '12345',
                     'vm_uuid': '12345a1',
-                    'state': 'Oversized',
-                    'recommendation_count': 4,
+                    'state': 'Crashloop',
```
apuntamb, line 17: Is `Oversized` no longer a state to be checked?

ros/resources/hosts.py
```diff
@@ -14,18 +14,18 @@ def get(self):
...
                     'organization_id': 1,
                     'performance_score': {
-                        'cpu_score': 70,
+                        'cpu_score': 20,
                         'memory_score': 20,
-                        'io_score': 80
+                        'io_score': 20
                     },
                     'facts': {
                         'provider': 'AWS',
```
patilsuraj767, line 26: I think `cloud_provider` will be more appropriate key here. Also, on host inventory side key name is `cloud_provider`

patilsuraj767 (changes requested):
LGTM :+1:  Just added one small point.

ros/resources/hosts.py
kgaikwad: State value should be based on scoring, right? I have modified the scoring for this record.

patilsuraj767 (approved):

apuntamb (approved):
looks ready! :+1:

apuntamb:
Merged, thanks @kgaikwad!

PR #8, opened 2020-12-24
Refs #18 Script building and pushing image to quay

patilsuraj767:


kgaikwad (commented):
@patilsuraj767,
Code wise changes looks good to me. 
When & where these changes will be used? I am assuming these are tasks that will be the part of some job.

apuntamb:
@patilsuraj767 Just a small query!
Looks like this will be helpful to push an initial image into the quay registry. Also, I guess if we link the push to our deployment config, it can trigger deployment run for us every time we push a new image.  But, what is the purpose of this script inside the repo? Can we not have it as a standalone script?

patilsuraj767:
> @patilsuraj767,
> Code wise changes looks good to me.
> When & where these changes will be used? I am assuming these are tasks that will be the part of some job.

Yes, this will be part of the app-interface job which will create and upload the image to quay repo.

patilsuraj767:
> @patilsuraj767 Just a small query!
> Looks like this will be helpful to push an initial image into the quay registry. Also, I guess if we link the push to our deployment config, it can trigger deployment run for us every time we push a new image. But, what is the purpose of this script inside the repo? Can we not have it as a standalone script?

@apuntamb there are two ways to create and upload the images to quay.  
First is through the Jenkins job on e2edeploy side. Here is the [PR](https://github.com/RedHatInsights/e2e-deploy/pull/2616/files#diff-a342406ef14523d7ecbac5cdf16f27a841fb2b3e60ff7d222cd8d57b58d91202R62) for it.  Second is through the app-interface job for which we need this PR. (Reference [code](https://github.com/RedHatInsights/insights-puptoo/blob/master/build_deploy.sh) from another project.)

kgaikwad (approved):
Thanks @patilsuraj767 for providing the details.

PR #10, opened 2021-01-04
ROS-21 v0 api desinging

apuntamb:
This PR handles the v0 api design for ros. It is currently in WIP state and requires discussion and review comments for merge ready. The flow of creating this PR is threefold as of now. Two ways that I could think of after reading various references are:

- either, using blueprints with view functions and registering the same to generate URLs for ros app and defining the api and api calls alongside of it: [read here](https://flask.palletsprojects.com/en/1.1.x/blueprints/#the-concept-of-blueprints).
- or, using just the project layout as defined by flask-restful: [read here](https://flask-restful.readthedocs.io/en/latest/intermediate-usage.html)

So, what are your opinions here, if we try to combine them together [like this](https://flask-restful.readthedocs.io/en/latest/intermediate-usage.html#use-with-blueprints) in order to, create our URLs to communicate to the host inventory similar to drift-backend: [read here](https://github.com/RedHatInsights/drift-backend/pull/10/files#diff-2061cd3e5999838b16eeb328da402476d11d0d17a6cd5d1ca44015ea5e8f49e5R1-R8) to do versionining on the API?

- [x] Remove existing dummy data
- [x] Fetch hosts lists from Inventory
- [ ] Fetch insights-client id to retrieve these hosts 
- [x] Define api calls/endpoints
- [x] Structure the api layout
- [x] Version this api v0

Please feel free to add comments if I'd missed any points.


apuntamb:
As next step, removed blueprints and added v0 api over resources layout.

apuntamb:
As next step, defined the api with realtime data.

Pipfile
```diff
@@ -17,6 +17,7 @@ flask-script = "~=2.0.6"
 flask-cors = "~=3.0.9"
 confluent-kafka = "~=1.5.0"
 requests = "~=2.25.1"
+marshmallow = "~=3.10.0"
```
patilsuraj767, line 20:
```suggestion

```

ros/resources/routes.py
```diff
@@ -0,0 +1,9 @@
+from .v0.hosts import HostsApi, HostDetailsApi
+from .status import Status
+
+
+# Initialize Routes
+def initialize_routes(api):
+    api.add_resource(Status, '/api/ros/v0/status')
+    api.add_resource(HostsApi, '/api/ros/v0/systems')
+    api.add_resource(HostDetailsApi, '/api/ros/v0/system/<host_id>')
```
patilsuraj767, line 9:
```suggestion
    api.add_resource(HostDetailsApi, '/api/ros/v0/systems/<host_id>')
```

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,67 @@
+import json
+from flask import jsonify
+from flask import request, Response
+from flask_restful import Resource, abort, fields, marshal_with
+from marshmallow import Schema, fields
```
patilsuraj767, line 5:
```suggestion
import uuid
```

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,67 @@
+import json
+from flask import jsonify
```
patilsuraj767, line 2:
```suggestion
from flask import jsonify make_response
```

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,67 @@
...
+from ros.models import PerformanceProfile
+from ros.utils import is_valid_uuid
+from ros.lib.host_inventory_interface import fetch_all_hosts_from_inventory
+
+import logging
+
+class PerformanceProfileSchema(Schema):
+    inventory_id = fields.Str()
+    performance_record = fields.Dict(keys=fields.Str(), values=fields.Str())
+    performance_score = fields.Dict(keys=fields.Str(), values=fields.Int())
```
patilsuraj767, line 16:
```suggestion

```

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,67 @@
...
+
+class HostsApi(Resource):
+    def get(self):
```
patilsuraj767, line 19:
```suggestion
class HostsApi(Resource):
    facts_fields = {
        'cloud_provider': fields.String,
        'instance_type': fields.String,
        'idling_time': fields.String,
        'io_wait': fields.String
    }
    performance_score_fields = {
        'cpu_score': fields.Integer,
        'memory_score': fields.Integer,
        'io_score': fields.Integer
    }
    hosts_fields = {
        'fqdn': fields.String,
        'display_name': fields.String,
        'id': fields.String,
        'account': fields.String,
        'recommendation_count': fields.Integer,
        'state': fields.String,
        'performance_score': fields.Nested(performance_score_fields),
        'facts': fields.Nested(facts_fields)

    }
    output_fields = {
        'results': fields.List(fields.Nested(hosts_fields))
    }

    @marshal_with(output_fields)
    def get(self):
```

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,67 @@
...
+        auth_key = request.headers.get('X-RH-IDENTITY')
```
patilsuraj767, line 20:
```suggestion
        auth_key = request.headers.get('X-RH-IDENTITY')
        if not auth_key:
            return make_response(
                jsonify({"Error": "Authentication token not provided"}), 401)
```

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,67 @@
...
+        inv_hosts.pop('count')
+        inv_hosts.pop('page')
+        inv_hosts.pop('per_page')
+
+        # Adding these fields as currently not present in rosdb
+        inv_hosts['results'][0]['recommendation_count'] = 5
+        inv_hosts['results'][0]['state'] = 'Crashloop'
+        print (inv_hosts)
+
+        return jsonify(inv_hosts)
```
patilsuraj767, line 51:
```suggestion
       inv_host_ids = [host['id'] for host in inv_hosts['results']]
        profile_result = PerformanceProfile.query.filter(PerformanceProfile.inventory_id.in_(inv_host_ids)).order_by(PerformanceProfile.report_date.desc()).all()
        
        hosts = []
        for i in profile_result:
            if len(list(filter(lambda host: host['id'] == str(i.inventory_id), hosts))):
                continue
            else:
                host = list(filter(lambda host: host['id'] == str(i.inventory_id), inv_hosts['results']))[0]
                host['performance_score'] = i.__dict__['performance_score']
                host['recommendation_count'] = 5
                host['state'] = 'Crashloop'
                hosts.append(host)

        return {"results": hosts}
```

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,67 @@
...
+
+class HostDetailsApi(Resource):
+    def get(self, host_id):
```
patilsuraj767, line 54:
```suggestion
class HostDetailsApi(Resource):
    profile_fields = {
        'host_id': fields.String(attribute='inventory_id'),
        'performance_record': fields.String,
        'performance_score': fields.String
    }

    @marshal_with(profile_fields)
    def get(self, host_id):
```

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,67 @@
...
+        if not is_valid_uuid(host_id):
+            abort(404, message='Invalid host_id,'
+                               ' Id should be in form of UUID4')
+
+        pp_schema = PerformanceProfileSchema()
+        profile = pp_schema.dump(PerformanceProfile.query.filter_by(
+                  inventory_id=host_id).first())
+        print (profile)
```
patilsuraj767, line 62:
```suggestion
        profile = PerformanceProfile.query.filter_by(
                  inventory_id=host_id).first()
```

patilsuraj767 (changes requested):
Is there really a need to introduce one more ORM `marshmallow`? We already have one i.e. `flask-sqlalchemy`.
Suggested few changes by which we can achieve the same results using flask-restful builtin decorator `@marshal_with`

ros/resources/v0/hosts.py
apuntamb: Doesn't it expect to have a model for the respective resource class we are defining these fields? For eg. in case of `fact_fields`, would it understand the 4 fields and consider them adding to the response as we do not have any model defined for the same or am I missing something here?

Pipfile
patilsuraj767: Ok, I didn't have an idea about `marshmallow` so I checked the home page [here](https://marshmallow.readthedocs.io/en/stable/index.html#) they mentioned it is an ORM so I compared it with SQLAlchemy

ros/resources/v0/hosts.py
apuntamb: @patilsuraj767 Wouldn't it be better to filter all inv_hosts whose ids match with profile.inventory_id instead of filtering profiles based on matching host_ids?
What do you think?

ros/resources/v0/hosts.py
patilsuraj767: > Doesn't it expect to have a model for the respective resource class we are defining these fields?

No, marshalling the object has no relation with the model. Consider the above fields as validation of our response. Explanation - the return object `hosts` of this function contains many unnecessary(sometimes sensitive info). Using `@marshal_with` only includes the fields in response which we have specified.

ros/resources/v0/hosts.py
patilsuraj767: Currently, we are not storing the account number of user in ROS db, so we cannot filter out the hosts from ROS db depending on the incoming request. I guess as per the current scenario we are first querying the host inventory using `X-RH-IDENTITY`(account number) and then filtering out those host from ROS DB. 
We can do the reverse way as you mentioned but then first we need to store account number with the help of ROS report-processor.

apuntamb: gotcha. also, in the suggestion why are we checking `len` on line 29?

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,80 @@
...
+        for i in profile_result:
+            if len(list(filter(lambda host: host['id'] == str(i.inventory_id), hosts))):
+                continue
+            else:
+                host = list(filter(lambda host: host['id'] == str(i.inventory_id), inv_hosts['results']))[0]
+                host['performance_score'] = i.__dict__['performance_score']
+                host['recommendation_count'] = 5
+                host['state'] = 'Crashloop'
+                hosts.append(host)
+        return {'results': hosts}
```
apuntamb, line 60: I'm currently seeing a `AttributeError: 'List' object has no attribute 'output'` on this GET call. I'll debug this and add changes further.

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,80 @@
...
+    output_fields = {
+        'results': fields.List(fields.Nested(hosts_fields))
+    }
+
+    @marshal_with(output_fields)
+    def get(self):
+        auth_key = request.headers.get('X-RH-IDENTITY')
+        if not auth_key:
+            return make_response(
+                jsonify({"Error": "Authentication token not provided"}), 401)
```
patilsuraj767, line 45:
```suggestion
            response = make_response(
                jsonify({"Error": "Authentication token not provided"}), 401)
            abort(response)
```

patilsuraj767 (changes requested):
Thank you @apuntamb for making above changes. Suggested one more below.

ros/resources/routes.py
patilsuraj767: Is it ok if we change API from `/api/ros/v0/system/<host_id>` to ` /api/ros/v0/systems/<host_id>`?

ros/exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+        
+        :type  status_code: int
+        
+        :param message:     optional error message to be put in the response
+                            body. If not supplied, the default message for the
+                            status code will be used.
+        """
+        super(HTTPError, self).__init__()
+        self.message = message
+        self.status_code = status_code
```
apuntamb, line 19: should I remove this class? I added it in case if we needed our exceptions, but doesn't look like we are using it.

patilsuraj767: Yes :+1:  better to remove if we are not using it.

ros/resources/v0/hosts.py
```diff
@@ -0,0 +1,81 @@
...
+        inv_host_ids = [host['id'] for host in inv_hosts['results']]
+        profile_result = PerformanceProfile.query.filter(
+            PerformanceProfile.inventory_id.in_(inv_host_ids)).order_by(
+                PerformanceProfile.report_date.desc()).all()
+        hosts = []
+        for i in profile_result:
+            if len(list(filter(lambda host: host['id'] == str(i.inventory_id), hosts))):
+                continue
+            else:
+                host = list(filter(lambda host: host['id'] == str(i.inventory_id), inv_hosts['results']))[0]
```
kgaikwad, line 54: Are we going to append inventory system detail with performance profile details?
I assumed exactly opposite to that as inventory host contains lot of other details that we don't required on ROS UI.

kgaikwad: @apuntamb,  as you are going to refactor this code in separate PR,  please ignore above comment.

kgaikwad:
@apuntamb @patilsuraj767 merging this PR.
Please make sure to address any unresolved comment(s) in separate PRs.

kgaikwad:
Thanks @apuntamb!

ros/resources/v0/hosts.py
apuntamb: Sure, thanks @kgaikwad

PR #11, opened 2021-01-07
ROS-20 Add .env.example file

apuntamb:
This PR adds a `.env.example` file for those users who wish to modify their ROS configurations. They can copy this template to a file called `.env` and ros-backend will start with these configs set.

### DIFF
--- a/ros/.env.example
+++ b/ros/.env.example
@@ -1,0 +1,11 @@
+ROS_DB_USER="postgres"
+ROS_DB_PASS="postgres"
+ROS_DB_HOST="localhost"
+ROS_DB_PORT="15432"
+ROS_DB_NAME="postgres"
+
+INVENTORY_HOST="localhost"
+INVENTORY_PORT="8081"
+
+INSIGHTS_KAFKA_HOST="localhost"
+INSIGHTS_KAFKA_PORT="29092"
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Afterwards you can activate the virtual environment by running:
 pipenv shell
 ```
 
+A list of configurable environment variables is present inside `.env.example` file.
+
 ## Initialize the database
 Run the following commands to excute the db migration scripts.
 ```bash


scripts/README.md
```diff
@@ -4,17 +4,22 @@ Below are the instructions to setup ROS development environment
...
 ```
 mkdir -p /usr/share/minio/{minio-conf,minio-data}
 ```
 
-3. Clone ros-backend repository and start necessary services in container.
+3. To make changes to ROS configurations, copy the example file and rename it as `.env` and update the environment variables.
+```
+cd ros; cp .env.example .env
+```
+
```
patilsuraj767, line 21: I think this should go in the main README file, as this step will be applicable in production too. 
Probably it can live in both the README files.

apuntamb: Sure, makes sense to have it in the main README.md

kgaikwad (changes requested):
@apuntamb, @patilsuraj767,
why `python-dotenv` package needed here? 
As per my understanding, this env file will be required ONLY for development or setup using docker.

@patilsuraj767, in case of openshift deployment(QA/PROD), using configmaps can't we set ENV variables?

patilsuraj767:
> @patilsuraj767, in case of openshift deployment(QA/PROD), using configmaps can't we set ENV variables?
Yes, In case of openshift we can set ENV variables using configmap. So no need of .env in case of PROD.

ros/.env.example
```diff
@@ -0,0 +1,5 @@
+ROS_DB_USER="postgres"
+ROS_DB_PASS="postgres"
+ROS_DB_HOST="localhost"
+ROS_DB_PORT="15432"
+ROS_DB_NAME="postgres"
```
patilsuraj767, line 5:
```suggestion
ROS_DB_NAME="postgres"

INVENTORY_HOST="localhost"
INVENTORY_PORT="8081"

INSIGHTS_KAFKA_HOST="localhost"
INSIGHTS_KAFKA_PORT="29092"
```

We should also add host-inventory and Kafka config here.

patilsuraj767 (changes requested):

apuntamb:
> @apuntamb, @patilsuraj767,
> why `python-dotenv` package needed here?
> As per my understanding, this env file will be required ONLY for development or setup using docker.

Okk sure makes sense! But, just a small query, how are the env variables then loaded for development if it's not set up using docker?
> 
> @patilsuraj767, in case of openshift deployment(QA/PROD), using configmaps can't we set ENV variables?

patilsuraj767:
> Okk sure makes sense! But, just a small query, how are the env variables then loaded for development if it's not set up using docker?

I think in the dev environment there is no need to load the env file as default values of the variables are set as per the dev environment.
Also, @kgaikwad need that env file just so that new user should know about configurable things/vars in ROS project. 
> > @patilsuraj767, in case of openshift deployment(QA/PROD), using configmaps can't we set ENV variables?

Yes, In case of openshift we can set ENV variables using configmap. So no need of .env in case of PROD.

kgaikwad (changes requested):
@apuntamb, 
I haven't seen any change in Pipefile so do we still need to commit `Pipfile.lock` file?

apuntamb:
> @apuntamb,
> I haven't seen any change in Pipefile so do we still need to commit `Pipfile.lock` file?


Sure thing! That was just an updated version of existing modules. Removed it. :)

patilsuraj767:
Thanks @apuntamb

PR #14, opened 2021-01-27
RHIROS-45 Create local-ros-backend.js for routes

apuntamb:
This route can be added in the frontend's local-frontend.js file as and when required for development purposes only.


kgaikwad (approved):
LGTM :+1:

PR #16, opened 2021-01-28
Update README.md

apuntamb:


README.md
```diff
@@ -28,18 +28,77 @@ python manage.py db upgrade
...
+`GET /api/ros/v0/systems` Shows list of all systems from Host Inventory having a Performance Profile
 
-To run via API:
-```
-curl -v -H "x-rh-identity: <identity-header-value>" -H "Content-Type: application/json" http://localhost:8080/api/ros/v0/systems
-curl -v -H "x-rh-identity: <identity-header-value>" -H "Content-Type: application/json" http://localhost:8080/api/ros/v0/status
-curl -v -H "x-rh-identity: <identity-header-value>" -H "Content-Type: application/json" http://localhost:8080/api/ros/v0/systems/<host_id>
-```
-For local dev setup, please note that use the x-rh-identity header encoded from your account number, the one used while running `make insights-upload-data` and `make ros-upload-data` commands.
+    curl -v -H "x-rh-identity: <header_value>" -H "Content-Type: application/json" https://ci.cloud.redhat.com/api/ros/v0/systems
```
patilsuraj767, line 36: If I am not wrong then I guess in CI env you will also need to pass username:password via `-u` option.

README.md
```diff
@@ -28,18 +28,77 @@ python manage.py db upgrade
...
+
+### Response
+
+    HTTP/1.1 200 OK
+    Date: Thu, 24 Feb 2011 12:36:30 GMT
+    Status: 200 OK
+    Connection: close
+    Content-Type: application/json
+    Content-Length: 2
```
patilsuraj767, line 45: I think we should remove this response header from the output/readme file.

patilsuraj767 (changes requested):

README.md
gnurag: I'd suggest using final production url in the examples: https://cloud.redhat.com/api/ros/v0/systems . Users trying out ci/qa/staging environments should be replacing it with their specific urls.

README.md
```diff
@@ -28,18 +28,77 @@ python manage.py db upgrade
...
+            'idling_time': '20',
+            'io_wait': '180'
+        }
+    }]
+
+
+### Request
+`GET /api/ros/v0/systems/<host_id>` Shows the Performance Profile of the selected <host_id>
+
+    curl -v -H "x-rh-identity: <header_value>" -H "Content-Type: application/json" https://ci.cloud.redhat.com/api/ros/v0/systems/<host_id>
```
gnurag, line 73: Same comment about removing `ci.` from the url.

README.md
```diff
@@ -28,18 +28,77 @@ python manage.py db upgrade
...
+    Content-Type: application/json
+    Content-Length: 2
+
+    {"host_id": "12345-57575757", "performance_record": "{'avg_memory': '3998008.000', 'avg_memory_used': '2487908.973'}", "performance_score": "{'memory_score': 62}"}
+
+
+### Request
+`GET /api/ros/v0/status` Shows the status of the server
+
+    curl -v -H "x-rh-identity: <header_value>" -H "Content-Type: application/json" https://ci.cloud.redhat.com/api/ros/v0/status
```
gnurag, line 90: Same comment about removing ci. from the url.

gnurag (changes requested):
Replace ci.cloud.redhat.com with cloud.redhat.com

README.md
apuntamb: Ok makes sense! So @patilsuraj767 would the request for prod require -u option too?

README.md
apuntamb: @patilsuraj767 I guess there should be no harm in showing the response header. For reference: https://github.com/bbc/REST-API-example/blob/master/README.md#create-a-new-thing. We can change it to precisely what details are shown in our response header. Thoughts?
@gnurag comments?

README.md
```diff
@@ -28,18 +28,77 @@ python manage.py db upgrade
...
+            'cloud_provider': 'AWS',
+            'instance_type': 'm4large',
+            'idling_time': '20',
+            'io_wait': '180'
+        }
+    }]
+
+
+### Request
+`GET /api/ros/v0/systems/<host_id>` Shows the Performance Profile of the selected <host_id>
```
kgaikwad, line 71:
```suggestion
`GET /api/ros/v0/systems/<host_id>`  to get the individual system detail using their <host_id>
```

API response should contain detailed information not just performance record & score. Makes sense?

apuntamb: Yeah sounds good!

README.md
patilsuraj767: Ok, we can keep the response header.

README.md
patilsuraj767: @apuntamb I am not sure regarding the prod env.

PR #18, opened 2021-02-01
implement pagination for systems list

kgaikwad:
Refer - https://platform-docs.cloud.paas.psi.redhat.com/rest/pagination.html

### DIFF
--- a/ros/resources/v0/hosts.py
+++ b/ros/resources/v0/hosts.py
@@ -5,36 +5,39 @@ from flask_restful import Resource, abort, fields, marshal_with
 from ros.models import PerformanceProfile
 from ros.utils import is_valid_uuid
 from ros.lib.host_inventory_interface import fetch_all_hosts_from_inventory
-
+from ros.api.common.pagination import build_paginated_system_list_response
 
 class HostsApi(Resource):
-    facts_fields = {
-        'cloud_provider': fields.String,
-        'instance_type': fields.String,
-        'idling_time': fields.String,
-        'io_wait': fields.String
-    }
-    performance_score_fields = {
-        'cpu_score': fields.Integer,
-        'memory_score': fields.Integer,
-        'io_score': fields.Integer
-    }
-    hosts_fields = {
-        'fqdn': fields.String,
-        'display_name': fields.String,
-        'id': fields.String,
-        'account': fields.String,
-        'recommendation_count': fields.Integer,
-        'state': fields.String,
-        'performance_score': fields.Nested(performance_score_fields),
-        'facts': fields.Nested(facts_fields)
-    }
-    output_fields = {
-        'results': fields.List(fields.Nested(hosts_fields))
-    }
+    # facts_fields = {
+    #     'cloud_provider': fields.String,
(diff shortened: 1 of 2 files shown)


ros/resources/v0/hosts.py
```diff
@@ -5,9 +5,14 @@
 from ros.models import PerformanceProfile
 from ros.utils import is_valid_uuid
 from ros.lib.host_inventory_interface import fetch_all_hosts_from_inventory
+from ros.api.common.pagination import build_paginated_system_list_response
+
+MAX_HOSTS_PER_REP = 1000
```
patilsuraj767, line 10: @kgaikwad outputting 1000 hosts per request seems to be very huge. Can we set the default limit to 10? 
Also, I think the variable name should be `DEFAULT_HOSTS_PER_REP`, As you are using it to set the `limit` value when `limit` is not passed in query params.

patilsuraj767 (changes requested):

ros/resources/v0/hosts.py
kgaikwad: makes sense :+1:

ros/resources/v0/hosts.py
```diff
@@ -43,20 +64,37 @@ def get(self):
...
+
+        # Note that When using LIMIT, it is important to use an ORDER BY clause
+        # that constrains the result rows into a unique order.
+        # Otherwise you will get an unpredictable subset of the query's rows.
+        # Refer - https://www.postgresql.org/docs/13/queries-limit.html
+        query = PerformanceProfile.query.filter(
             PerformanceProfile.inventory_id.in_(inv_host_ids)).order_by(
-                PerformanceProfile.report_date.desc()).all()
+                PerformanceProfile.report_date.desc()).order_by(
+                    PerformanceProfile.id.asc())
```
patilsuraj767, line 75: Adding the `asc()` order_by clause will change the ordering/result and it can break the below logic of getting distinct hosts with latest report_date.

kgaikwad: @patilsuraj767, @apuntamb,
Yes! Currently I have applied order_by `id`  because I was getting an unpredictable rows & few rows were missing from DB.
`report_date` values can be same for records so not sure if that gives unique order.

As per my understanding, fetching of unique host_id's latest record logic should be handled at query level not here ->
https://github.com/RedHatInsights/ros-backend/pull/18/files#diff-e303aa4e3e61a69895b0d295ce9166a08965a63a0c63569a4e1c8c60235f3b50R82

I have reported a bug  - https://issues.redhat.com/browse/RHIROS-59. 
This change should not be a part of pagination change.

kgaikwad: I have marked issue ticket as bug because currently I applied order by `id` but for sorting change, this implementation needs to be fixed.

patilsuraj767 (approved):

apuntamb (approved):
Tested locally, works as expected.

apuntamb:
Merged, thanks @kgaikwad and @patilsuraj767 !

PR #20, opened 2021-02-04
ROS-26 Compute scores in API

apuntamb:
Computing scores `1-5` in the model.

### DIFF
--- a/ros/resources/v0/hosts.py
+++ b/ros/resources/v0/hosts.py
@@ -88,6 +88,9 @@ class HostsApi(Resource):
                     {'cpu_score': 60, 'io_score': 30}
                 )
                 host['id'] = i.__dict__['id']
+                host['performance_score']['memory_score'] //= 20
+                host['performance_score']['cpu_score'] //= 20
+                host['performance_score']['io_score'] //= 20
                 host['recommendation_count'] = 5
                 host['state'] = 'Undersized'
                 hosts.append(host)
@@ -112,7 +115,13 @@ class HostDetailsApi(Resource):
 
         profile = PerformanceProfile.query.filter_by(
                   inventory_id=host_id).first()
-        if not profile:
+        if profile:
+            p_score_calc = profile.performance_score
+            p_score_calc['memory_score'] //= 20
+            p_score_calc['cpu_score'] //= 20
+            p_score_calc['io_score'] //= 20
+
+        else:
             abort(404, message="Performance Profile {} doesn't exist"
                   .format(host_id))
 
--- a/ros/resources/v0/hosts.py
+++ b/ros/resources/v0/hosts.py
@@ -19,7 +19,7 @@ class HostsApi(Resource):
         'idling_time': fields.String,
         'io_wait': fields.String
     }
-    performance_score_fields = {
+    display_performance_score_fields = {
         'cpu_score': fields.Integer,
         'memory_score': fields.Integer,
         'io_score': fields.Integer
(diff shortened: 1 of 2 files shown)


ros/resources/v0/hosts.py
```diff
@@ -53,6 +53,9 @@ def get(self):
             else:
                 host = list(filter(lambda host: host['id'] == str(i.inventory_id), inv_hosts['results']))[0]
                 host['performance_score'] = i.__dict__['performance_score']
+                host['performance_score']['memory_score'] //= 20
```
kgaikwad, line 56: Will it be possible to move or add properties at model level?

apuntamb: @semtexzv any thoughts here?

kgaikwad: With `@property` also it is possible to define custom field at model level.
example -

    @property
    def score(self):
        print("Getting value...")
        return 3

the issue was due to `marshal_with`. In this fields, forgot to declare field name and type for new property.
https://github.com/RedHatInsights/ros-backend/blob/main/ros/resources/v0/hosts.py#L27after adding new property details under hosts_fields, it is giving property details in response as well.

@apuntamb, could you please make changes and test it locally if that works for you.

patilsuraj767: +1 for @kgaikwad suggestion of having `@property ` inside model.

ros/resources/v0/hosts.py
```diff
@@ -90,6 +91,8 @@ def get(self):
                 host['id'] = i.__dict__['id']
```
kgaikwad, line 91: From documentation, `object.__dict__` is A dictionary or other mapping object used to store an object’s (writable) attributes.
That's  why you are not getting new property inside `__dict__`.
Could you please change the way you are accessing the attributes to fix this?

ros/resources/v0/hosts.py
```diff
@@ -90,6 +91,8 @@ def get(self):
...
                 host['recommendation_count'] = 5
                 host['state'] = 'Undersized'
+                pro = PerformanceProfile()
+                pro.my_score
```
kgaikwad, line 95: You already have performance_profile instance (here it's `i`) then why can't you use it directly instead of calling `PeroformanceProfile`?
Example - `i.my_score`

kgaikwad: Also, instead `i` , will it be good to rename this variable to something else so that one can get what information is `stored` inside that variable

kgaikwad, line 95:
```suggestion
                host['my_score'] = i.my_score
```
With this, you will get value inside response.

kgaikwad (changes requested):
Added few inline comments. 
I hope those will be helpful and you will get `my_score` inside response.

ros/resources/v0/hosts.py
```diff
@@ -31,7 +31,7 @@ class HostsApi(Resource):
         'account': fields.String,
         'recommendation_count': fields.Integer,
         'state': fields.String,
-        'performance_score': fields.Nested(performance_score_fields),
+        'display_performance_score': fields.Nested(display_performance_score_fields),
```
apuntamb, line 34: Thanks for the suggestions @kgaikwad, made all the changes accordingly.
But, there is still one thing that I'm guessing, since the field name is not `performance_profile` now, it doesn't show the records on UI.
Anything specific about this or I'm missing something here? Could you test it locally on UI?

kgaikwad (approved):
Looks good now. :+1: :ok_hand: 
Thanks @apuntamb! Merging it

PR #21, opened 2021-02-08
Fix filtering of hosts

patilsuraj767:

PR #22, opened 2021-02-08
Fix development setup issues.

semtexzv:
- Instead of creating minio directories within the system directories of local machine, create them in the project directory.
- There is a port conflict between ROS backend and inventory containers
- Allow connections to kafka for both `container -> container` and `host -> container` interactions.


run.py
```diff
@@ -1,4 +1,4 @@
 from ros.app import app
 
 if __name__ == "__main__":
-    app.run(host='0.0.0.0', port=8080)
+    app.run(host='0.0.0.0', port=8082)
```
patilsuraj767, line 4: We use the same run.py inside Dockerfile in `CMD` Also, 3scale service redirects the request to port 8080. 
I think we should consider changing the port of ingress container in docker-compose. I have raised the PR here - [https://github.com/RedHatInsights/ros-backend/pull/21/files](https://github.com/RedHatInsights/ros-backend/pull/21/files)
But this PR is a better fit for those changes.

patilsuraj767 (changes requested):
Thank for the PR @semtexzv 
Add a small concern below.

run.py
semtexzv: Thank you, I'll either integrate changes from that PR here, or just wait for them to be merged.

patilsuraj767: +1 to make those changes here. As this PR is more relevant for those changes.

scripts/.gitignore
```diff
@@ -0,0 +1,2 @@
+./minio-conf/
+./minio-data/
```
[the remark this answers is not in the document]
semtexzv: Thank you, fixed.

patilsuraj767 (approved):
Thank you @semtexzv  :+1: 
Merging the PR.

PR #27, opened 2021-02-11
Dev setup improvments

semtexzv:
Small changes to dev setup
- Custom identity
- Run both inventory MQ and Web services
- Should allow better end to end testing


kgaikwad (approved):
Tested it, working as expected. :+1:

scripts/docker-compose.yml
```diff
@@ -79,38 +80,69 @@ services:
...
+    build:
+      context: ..
+    command: [ "python", "-m", "ros.processor.main" ]
+    depends_on:
+      - insights-inventory-web
+      - kafka
+      - createbuckets
+    environment:
+      - INSIGHTS_KAFKA_HOST=kafka
+      - INSIGHTS_KAFKA_PORT=29092
```
kgaikwad, line 141:
```suggestion
      - INSIGHTS_KAFKA_PORT=9092
```
As per previous changes made, in the ROS config file & here, modified port to `9092`. At this line also will it be good to have `9092`?

semtexzv: When running inside docker, we need to set different port. The kafka container listens on both, but kafka consists of several parts. What we contact at first is the broker

Dockerfile
```diff
@@ -1,8 +1,9 @@
 FROM registry.redhat.io/rhscl/python-38-rhel7
+
 COPY Pipfile Pipfile.lock manage.py run.py ${APP_ROOT}/src/
+
 COPY ros ${APP_ROOT}/src/ros/
 COPY migrations ${APP_ROOT}/src/migrations/
 RUN pip install --upgrade pip && \
     pip install pipenv && \
     pipenv install --system --deploy --ignore-pipfile
-CMD bash -c 'python manage.py db upgrade && python run.py'
```
patilsuraj767, line 8: If we are planning to remove `CMD`. Then there should be a readme which tells how to run the container using docker. I mean we will need to pass `CMD` in `docker run` command explicitly. Also, we will need to add [command and args ](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/)in e2edeploy so that when Openshift spins up the container there should be an initial process to run.

.dockerignore
```diff
@@ -0,0 +1,2 @@
+scripts/minio-conf/
+scripts/minio-data/
```
patilsuraj767, line 2: As this PR is related to minio directory changes. Can you please remove this line - [here](https://github.com/RedHatInsights/ros-backend/blame/main/scripts/README.md#L14) from the Readme

scripts/docker-compose.yml
```diff
@@ -79,38 +80,69 @@ services:
...
+      - INVENTORY_HOST=insights-inventory-web
+      - INVENTORY_PORT=8080
+      - ROS_DB_HOST=db-ros
+      - ROS_DB_PORT=5432
+  ros-api:
```
patilsuraj767, line 146: If I am not wrong then while development we will need to manually stop the `ros-api` service/container right?

scripts/docker-compose.yml
```diff
@@ -79,38 +80,69 @@ services:
...
       - PAYLOAD_TRACKER_ENABLED=false
       - LISTEN_PORT=8081
     depends_on:
        - kafka
        - db-host-inventory
        - puptoo
-       - zookeeper
+
+  insights-inventory-web:
+    <<: *inventory
```
patilsuraj767, line 107: I didn't get the idea of running inventory container two times, I can also see that migration is also running twice.

patilsuraj767 (changes requested):

PR #29, opened 2021-02-15
ROS-65 Create new tables in DB

apuntamb:
Adding new tables `systems`, `reports` and `accounts` to store selected data for the systems which use ros, instead of calling inventory API.


ros/models.py
```diff
@@ -25,3 +25,30 @@ def display_performance_score(self):
...
+    __table_args__ = (
+        db.ForeignKeyConstraint(['system_id'], ['systems.id']),
+        db.PrimaryKeyConstraint('system_id', 'reported'),
+    )
+
+
+class RhAccount(db.Model):
+    __tablename__ = 'rh_accounts'
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.Text, nullable=False)
```
apuntamb, line 51: @semtexzv should we have `account_number` column instead of `name` here?

semtexzv: Probably a better name, yes.

migrations/versions/a41fc890a5a8_.py
```diff
@@ -0,0 +1,58 @@
+"""empty message
+
+Revision ID: a41fc890a5a8
```
semtexzv, line 3: Was the application deployed to stage environment ? If it was only in QA + CI , we can reset those databases without problems, but for other environments, we need to create a proper migration which will either remove old `performance_profile` table or alter it.

apuntamb: Actually, there are no new changes in `performance_profile` table. So, should this [file](https://github.com/RedHatInsights/ros-backend/pull/29/files/b46fd995a313a15c6015baf31ed7a0945290f9d9#diff-f5e6df0d16ec0fcbf5c876f937ee6b8222043593b875177c50dcd513dc480af2) remain as is and a new migration only for new entities added?

migrations/versions/a41fc890a5a8_.py
```diff
@@ -0,0 +1,58 @@
...
+    op.create_table('systems',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('account_id', sa.Integer(), nullable=True),
+    sa.Column('inventory_id', postgresql.UUID(as_uuid=True), nullable=False),
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('inventory_id')
+    )
+    op.create_table('reports',
+    sa.Column('system_id', sa.Integer(), nullable=False),
+    sa.Column('reported', sa.Date(), nullable=False),
```
semtexzv, line 45: We can just alter the performance_profile table to contain a foreign key into the `systems` table.

migrations/versions/a41fc890a5a8_.py
semtexzv: Yes, in that case the old migration should remain. We need to have an unbroken chain of modifications to the database that will result in final schema.

migrations/versions/a41fc890a5a8_.py
apuntamb: done.

migrations/versions/a41fc890a5a8_.py
apuntamb: done.

ros/models.py
apuntamb: done.

ros/models.py
```diff
@@ -6,12 +6,16 @@
 
 
 class PerformanceProfile(db.Model):
-    __table_args__ = (db.UniqueConstraint('inventory_id', 'report_date'), )
     id = db.Column(db.Integer, primary_key=True)
     inventory_id = db.Column(UUID(as_uuid=True), nullable=False)
```
semtexzv, line 10: We can remove this column, since we have it in system table. Otherwise looks good :+1: .

apuntamb: Makes sense, but, removing this would affect the API call `/systems` [here](https://github.com/RedHatInsights/ros-backend/blob/main/ros/resources/v0/hosts.py#L73-L86). So, this should not be merged until the API call is fixed.
Thoughts?

semtexzv: Ok, then I think this can be merged

apuntamb: You mean merged with `inventory_id` present in `performance_profile` table, or without?

semtexzv: Removed, the app will still be nonfunctional until we implement changes in the processor and the API.

kgaikwad:
Any additional comments/suggestions @semtexzv, @patilsuraj767 ?

ros/models.py
```diff
@@ -6,12 +6,15 @@
...
 class PerformanceProfile(db.Model):
-    __table_args__ = (db.UniqueConstraint('inventory_id', 'report_date'), )
     id = db.Column(db.Integer, primary_key=True)
-    inventory_id = db.Column(UUID(as_uuid=True), nullable=False)
     performance_record = db.Column(JSONB)
     performance_score = db.Column(JSONB)
     report_date = db.Column(db.Date, default=date.today())
+    system_id = db.Column(db.Integer, primary_key=True)
+    __table_args__ = (
+        db.UniqueConstraint('report_date'),
```
kgaikwad, line 15: More than one system can report on same day then will it work by declaring unique_key only on `report_date` column? 
It should consider single unique_key using both `report_date` & `system_id`columns.

semtexzv: Oh I must've missed that. The primary key should be `(system_id,report_date)`

semtexzv: Yeah, @apuntamb is working on changing this table to use `system_id, report_date` as a primary key.

ros/models.py
```diff
@@ -6,12 +6,15 @@
 
 
 class PerformanceProfile(db.Model):
-    __table_args__ = (db.UniqueConstraint('inventory_id', 'report_date'), )
     id = db.Column(db.Integer, primary_key=True)
-    inventory_id = db.Column(UUID(as_uuid=True), nullable=False)
     performance_record = db.Column(JSONB)
     performance_score = db.Column(JSONB)
     report_date = db.Column(db.Date, default=date.today())
+    system_id = db.Column(db.Integer, primary_key=True)
```
kgaikwad, line 13: @apuntamb, do we need to declare `system_id` column as primary_key?
With this change, will you be able to store historical records?

apuntamb: To store historical records at a later stage, we can create a new table something like `historical_pp` that can have system_id as the foreign key and store those multiple records with unique dates, does that sound ok?

kgaikwad: @apuntamb, why do we need `system_id` as primary_key in `PerformanceProfile`?
I think having primary_key on  `(system_id, report_date)` serving the purpose. Sorry but I didn't get the point of creating a separate table `historical_pp` as it will be nothing but the copy of `PerformanceProfile` table.

ros/models.py
```diff
@@ -6,12 +6,15 @@
 
 
 class PerformanceProfile(db.Model):
-    __table_args__ = (db.UniqueConstraint('inventory_id', 'report_date'), )
     id = db.Column(db.Integer, primary_key=True)
```
semtexzv, line 9: Now that we use system_id, we don't have to have the id column

ros/models.py
semtexzv: We have a separate PrimaryKeyConstraint specified, so the system_id does not have to have the primary_key annotation

ros/models.py
```diff
@@ -25,3 +28,20 @@ def display_performance_score(self):
             # 'io_score': self.performance_score['io_score'] // 20
         }
         return display_performance_score
+
+
+class System(db.Model):
+    __tablename__ = 'systems'
+    id = db.Column(db.Integer, primary_key=True)
+    account_id = db.Column(db.Integer)
```
semtexzv, line 36: Missing ForeignKeyConstraint

semtexzv (approved):

kgaikwad (approved):
Thanks @apuntamb & @semtexzv :+1: 
Merging it

kgaikwad:
As everything from `main` branch is pushed to `stable` branch. Refer link - https://github.com/RedHatInsights/ros-backend/pull/32#pullrequestreview-592938537
Merging this branch.

PR #31, opened 2021-02-17
Handle Create/Update messages

semtexzv:
Blocked on #29


ros/processor/inventory_events_consumer.py
```diff
@@ -79,13 +86,45 @@ def host_delete_event(self, msg):
                 insights_id,
                 self.prefix
             )
-            rows_deleted = db.session.query(PerformanceProfile).filter(
-                PerformanceProfile.inventory_id == host_id
-            ).delete()
+            system_id_query = db.session.query(System.id).filter(System.inventory_id == host_id).subquery()
+            rows_deleted = db.session.query(PerformanceProfile) \
+                .filter(PerformanceProfile.system_id.in_(system_id_query))
```
patilsuraj767, line 91: I think you missed to chain the `.delete()` function here.

ros/processor/inventory_events_consumer.py
```diff
@@ -79,13 +86,45 @@ def host_delete_event(self, msg):
...
+        """ Store new system information (stale, stale_warning timestamp) and return internal DB id"""
+        host = msg['host']
+
+        account = get_or_create(
+            db.session, RhAccount, 'account',
+            account=host['account']
+        )
+
+        db.session.add(account)
+        db.session.flush()
```
patilsuraj767, line 118: Here you might end up trying to save account two times in DB. Because in `get_or_create` you are already doing `session.add(instance)` and `session.flush()`

patilsuraj767 (changes requested):
I have not tested the PR, But added few things which I noticed at first glance.

ros/lib/utils.py
```diff
@@ -7,3 +7,21 @@ def is_valid_uuid(val):
...
+
+
+def get_or_create(session, model, keys, **kwargs):
+    if not keys:
+        keys = kwargs.keys()
+    if isinstance(keys, str):
+        keys = [keys]
+    if not isinstance(keys, list):
+        raise TypeError('keys argument must be a list or string')
+    instance = session.query(model).filter_by({k: kwargs[k] for k in keys}).first()
```
patilsuraj767, line 19: Getting below error on this line. 
`ERROR  - run - An error occurred during message processing: TypeError('filter_by() takes 1 positional argument but 2 were given')`

patilsuraj767 (approved):
I have a concern regarding the delete event code that it will not work as expected. 
But as its not part of this PR, Approving the PR :+1: 
Thank you @semtexzv :smiley:

PR #33, opened 2021-02-17
Fix openapi path

semtexzv:


apuntamb:
Thanks @semtexzv for the quick fix, merged!

PR #35, opened 2021-02-23
Add New fields to systems schema

patilsuraj767:


ros/lib/models.py
```diff
@@ -34,6 +34,11 @@ class System(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     account_id = db.Column(db.Integer)
     inventory_id = db.Column(UUID(as_uuid=True), nullable=False)
+    display_name = db.Column(db.String(100))
```
apuntamb, line 37: do we want the display name to be unique?

patilsuraj767: No, it may happen that two user account has the system with the same display name. So it cannot be unique.

apuntamb:
I guess the PR is already clean to get in. Just one query: do we also need to store Idling Time and IO wait fields in the `systems` table (fields shown in [mockup](https://marvelapp.com/prototype/9408d9j/screen/74816603))?

patilsuraj767:
> I guess the PR is already clean to get in. Just one query: do we also need to store Idling Time and IO wait fields in the `systems` table (fields shown in [mockup](https://marvelapp.com/prototype/9408d9j/screen/74816603))?

We are not getting the idling time and IO wait from inventory, I think we will need to find a way to populate it through PCP metric. So I think the better place to store would be the performance profile table under performance record.

apuntamb (approved):
LGTM :+1:

apuntamb:
Clean fix.
Merged, thanks @patilsuraj767!

PR #37, opened 2021-02-24
DB schema & API for recommendation rating 

kgaikwad:
Pending work:
- [x] openapi file needs to be updated with new API details
- [x] handle validation errors.


ros/lib/utils.py
```diff
@@ -25,3 +26,18 @@ def get_or_create(session, model, keys, **kwargs):
         session.add(instance)
         session.flush()
     return instance
+
+
+def identity(request):
```
kgaikwad, line 31: I will remove this change once original PR get merged.

patilsuraj767 (approved):
Thank you for the PR @kgaikwad 
Tested it on local dev setup, APIs are working as expected. Also, no issues in migration. 

@anurag03 @mayurilahane is it possible to write a smoke/integration test for testing migrations, 
In local dev env, we mostly make use of clean/fresh DB(without any record) while reviewing the PR. Due to this in past, in one of ROS migration, I faced an issue where migration failed to upgrade DB in CI env due to implementing `nullable=False`. 

Test scenario/steps - 
1) Run ROS from the master branch. 
2) Make few entries / upload few reports
3) Pull the new PR and run ROS using the same previous DB used in step 1 and 2. 
4) Make sure step 3 did not fail with migration error. 

PS - Smoke test is not a requirement for this PR to get merge, above is just an idea to have a test case to avoid migration error  /downtime when ROS will be in PROD.

PR #38, opened 2021-02-25
Modifed report processor to process insights archive

patilsuraj767:


ros/processor/inventory_events_consumer.py
```diff
@@ -103,27 +108,79 @@ def host_create_update_events(self, msg):
...
+
+    def _download_report(self, report_url):
+        download_response = requests.get(report_url)
+        if download_response.status_code != HTTPStatus.OK:
+            print("Unable to download the report")
+        return download_response.content
+
+    def _extract_performance_record(self, report_tar):
+        tar = tarfile.open(fileobj=BytesIO(report_tar), mode='r:*')
+        files = tar.getmembers()
```
kgaikwad, line 157: Instead of `getmembers` and below iteration on files, will it be possible to make use of `getmember` and fetch whichever required?
https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.getmember

ros/processor/inventory_events_consumer.py
```diff
@@ -103,27 +108,79 @@ def host_create_update_events(self, msg):
...
+        performance_record = {}
+        spec_count = 0
+        for file in files:
+            if 'data/insights_commands/pmlogsummary' in file.name:
+                spec_count += 1
+                extracted_pmlogsummary_file = tar.extractfile(file)
+                record_string = extracted_pmlogsummary_file.read().decode('utf-8')
+                records = record_string.split('\n')
```
kgaikwad, line 165: It seems similar logic needed for each file read, it will be better to extract this in separate method & use it.

kgaikwad (changes requested):
Hello @patilsuraj767, thank you for processor changes.
Added few inline comments, please have a look and let me know if any concerns regarding those.

ros/processor/inventory_events_consumer.py
patilsuraj767: Using `getmember` will be bit hard as tar files contain the hostname of the client system, So for using `getmember` first we will need to find out the hostname of the system. Also, I not sure if tar will always be having hostname in the file path. 
```
>>> tar.getmember('./insights-ip-172-31-0-61.ap-south-1.compute.internal-20210212124759/data/insights_commands/pmlogsummary')
<TarInfo './insights-ip-172-31-0-61.ap-south-1.compute.internal-20210212124759/data/insights_commands/pmlogsummary' at 0x7f67e41cf1c0>
```
Let keep this for now, we can fix it later as a scope of performance improvement.

ros/processor/inventory_events_consumer.py
```diff
@@ -118,38 +117,38 @@ def process_system_details(self, msg):
...
-                performance_record=performance_record,
-                performance_score=performance_score,
-                report_date=date.today()
-            )
+                get_or_create(
+                    db.session, PerformanceProfile, ['system_id', 'report_date'],
+                    system_id=system.id,
+                    performance_record=performance_record,
+                    performance_score=performance_score,
+                    report_date=date.today()
```
kgaikwad, line 140: Do we need to take date after converting it to `utc`?

kgaikwad (approved):
As puptoo changes are not yet merged, with replacing `is_ros` to `url` at [line](https://github.com/RedHatInsights/ros-backend/pull/38/files#diff-3ec10f47264a06a134d0bf1ce745fb26ad086de0af6da0cf6e83c00dcf2c4193R110) ,
I am able to see the changes in all tables created in ROS & getting profile details on making `/systems` request.

Note - Once everything available to support`v2` architecture fully, in future we don't need `ros-upload-data` make command so we can remove it later.

kgaikwad:
Great work @patilsuraj767 :ok_hand: :ice_cream:

PR #42, opened 2021-03-03
Implement insights core inside ROS

patilsuraj767:
With this PR, ROS processor will be using insights-core spec and parsers for extracting performance profile data from tar archive

- [x] Implement insights core in ros
- [x] Change AWS sample archive


ros/processor/process_archive.py
```diff
@@ -0,0 +1,66 @@
...
+        'kernel.all.cpu.idle',
+        'disk.all.total',
+        'mem.util.cached',
+        'mem.util.bufmem',
+        'mem.util.free'
+        ]
+    profile["total_cpus"] = int(lscpu.info.get('CPUs'))
+    profile["instance_type"] = aws_instance_id.get('instanceType')
+    for i in performance_metrics:
+        profile[i] = _.get(pmlog_summary, f'{i}.val', default='None')
```
kgaikwad, line 38:
```suggestion
        profile[i] = _.get(pmlog_summary, f'{i}.val')
```

default value is `None` only, right? if yes, will it be ok to pass only object & path?

ros/processor/process_archive.py
```diff
@@ -0,0 +1,66 @@
...
+    metadata_response.update(profile)
+    return metadata_response
+
+
+def get_performance_profile(report_url):
+    with _download_and_extract_report(report_url) as archive:
+        try:
+            broker = run(performance_profile, root=archive.tmp_dir)
+            result = broker[performance_profile]
+            del result["type"]
```
kgaikwad, line 50: What information this `type` key holds?

kgaikwad (changes requested):
Small inline comments added. 
Rest code looks good but will it be good to request someone who is more familiar with insights helpers that used here to get profile details?

ros/processor/process_archive.py
patilsuraj767: `{'type': 'metadata', 'total_cpus': 1, 'instance_type': 't2.micro', 'mem.physmem': 825152.0, 'mem.util.used': 663245.405, 'kernel.all.cpu.user': 0.003, 'kernel.all.cpu.sys': 0.001, 'kernel.all.cpu.nice': 0.001, 'kernel.all.cpu.steal': 0.0, 'kernel.all.cpu.idle': 0.994, 'disk.all.total': 2.727, 'mem.util.cached': 351730.563, 'mem.util.bufmem': 335.552, 'mem.util.free': 161906.595}`

kgaikwad:
Thanks @patilsuraj767

PR #46, opened 2021-03-10
Process report from insights-engine result topic

patilsuraj767:


ros/processor/insights_engine_result_consumer.py
```diff
@@ -0,0 +1,90 @@
...
+                    repr(err),
+                    self.prefix
+                )
+            finally:
+                self.consumer.commit()
+
+    def handle_msg(self, msg):
+        if msg["input"]["platform_metadata"]["is_ros"]:
+            host = msg["input"]["host"]
+            reports = msg["results"]["reports"]
```
kgaikwad, line 63: @patilsuraj767, could you please share sample `report` detail here that would be helpful while reviewing this PR.

patilsuraj767: @kgaikwad added the sample file `insights-engine-result.json ` which contains the payload from `platform.engine.results `
In `insights-engine-result.json ` you will also find the ros rule_hit in the reports.  `"rule_id": "cloud_instance_ros_evaluation|CONSUMPTION_MODEL"`

ros/processor/insights_engine_result_consumer.py
```diff
@@ -0,0 +1,90 @@
...
+                    account_id=account.id,
+                    inventory_id=host['id'],
+                    display_name=host['display_name'],
+                    fqdn=host['fqdn'],
+                    rule_hit_details=reports
+                )
+
+            db.session.commit()
+            LOG.info("Refreshed system %s (%s) belonging to account: %s (%s) via engine-result",
+                     system.inventory_id, system.id, account.account, account.id)
```
apuntamb, line 90: I didn't understand why do we need to `get_or_create` System or RHAccount in both consumers?

patilsuraj767: It is not sure which consumer will consume the report first, It might happen that `InsightsEngineResultConsumer` will consume the message before the `InventoryEventsConsumer`. So instead of ignoring the data coming from `InsightsEngineResultConsumer` due to no base record present in the database, it is better to create the new record from available data. And later `InventoryEventsConsumer` will populate the remaining record/data.

apuntamb: So, why populate remaining data from `InventoryEventsConsumer` in that case? Can we create a complete record (data from single report) in each of the tables from a single consumer? By this I mean, we can `get_or_create` for all the tables in both consumers. Also, to avoid repetition of same code blocks, can we just define a single function and use it in both the consumers under the app_context?

patilsuraj767: Yes, we can do that I will update the PR.

migrations/versions/7805dfde3ca9_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '7805dfde3ca9'
+down_revision = 'ac76a3ecb338'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('systems', sa.Column('rule_hit_details', postgresql.JSONB(astext_type=sa.Text()), nullable=True))
```
kgaikwad, line 21: Will it be good to have a separate table to store rule hit details?
i.e with columns - id, system_id, account_id, rule_id, rule_hit_details

patilsuraj767: Hello, what should be the rule_id? looking at [this](https://github.com/RedHatInsights/ros-backend/pull/46/files#diff-a0e4b82fb8caf907a6fdd12040acaeea35910491cb3e7a7c038e99b94d809822R228), I think rule_id will be dynamic and only part "cloud_instance_ros_evaluation" will be constant.

ros/lib/config.py
```diff
@@ -15,6 +15,7 @@
 INVENTORY_URL = f"http://{INVENTORY_HOST}:{INVENTORY_PORT}"
 INVENTORY_EVENTS_TOPIC = os.getenv("INVENTORY_EVENTS_TOPIC", "platform.inventory.events")
 
+Engine_RESULT_TOPIC = os.getenv("INVENTORY_EVENTS_TOPIC", "platform.engine.results")
```
kgaikwad, line 18:
```suggestion
Engine_RESULT_TOPIC = os.getenv("Engine_RESULT_TOPIC", "platform.engine.results")
```

migrations/versions/7805dfde3ca9_.py
kgaikwad: ahh okies! For every report, it has different `rule_id` &  you are storing all reports in one column, right? I think just ignore `rule_id` for now.

ros/lib/config.py
apuntamb: Just a minor nitpick if we can make it:
```suggestion
ENGINE_RESULT_TOPIC = os.getenv("ENGINE_RESULT_TOPIC", "platform.engine.results")
```

apuntamb (approved):
@patilsuraj767 @kgaikwad tested the PR end-to-end.
Current workflow in the local developmemt environment:
1. `cat <wrapped-json file> | docker-compose exec -T kafka kafka-console-producer --topic platform.engine.results --broker-list localhost:29092` to populate systems and rh_account record
2. `make insights-upload-data` to populate performance_profile record.

Works fine. We can consider solving non-specifically related issues in separate PR.

ros/lib/config.py
kgaikwad: Thanks @apuntamb  for correction :)

kgaikwad (approved):

kgaikwad:
Thank you @patilsuraj767 :ok_hand:  :+1: !
Thanks @apuntamb for review & testing the PR.

PR #49, opened 2021-03-16
sort systems list by display_name & score fields

kgaikwad:


ros/api/v0/hosts.py
```diff
@@ -102,6 +111,48 @@ def get(self):
...
+            return (
+                sort_order(PerformanceProfile.performance_score[
+                    'memory_score'].astext.cast(Integer)),
+                asc(PerformanceProfile.system_id), )
+        if order_method == 'io_score':
+            return (
+                sort_order(PerformanceProfile.performance_score[
+                    'io_score'].astext.cast(Integer)),
+                asc(PerformanceProfile.system_id), )
+        # FIXME: no ordering as of now for columns:
```
apuntamb, line 151: For the scores as order_methods, can we do something like:
```python
score_methods = ['memory_score', 'cpu_score', 'io_score']
if order_method in score_methods:
    return return (
                sort_order(PerformanceProfile.performance_score[
                    order_method].astext.cast(Integer)),
                asc(PerformanceProfile.system_id)
            )
```

kgaikwad: Good Catch :+1:

apuntamb:
@kgaikwad, works perfect with individual params, but when I combine the two eg. `?order_by=memory_score&order_how=DESC`, it doesn't. Could you check if the same is happening in your local?

patilsuraj767:
Tested the PR locally. LGTM :+1: 
@apuntamb regarding `?order_by=memory_score&order_how=DESC` make sure you surround the API in curl command with double quotes

apuntamb (approved):
LGTM! :+1:

apuntamb:
Merged, thanks @kgaikwad !! :+1:

PR #51, opened 2021-03-16
RHIROS-44 Filter systems with individual scores zero

apuntamb:
To handle showing only those systems which do not have any of the scores as 0.


apuntamb:
@patilsuraj767 Do we still want this PR now that you've created #60 ?

patilsuraj767:
@apuntamb I don't think we require it anymore, As we will be setting the score default to 1.

PR #52, opened 2021-03-16
Fix KeyError for rule_hit_details

patilsuraj767:


ros/api/v0/hosts.py
```diff
@@ -92,8 +92,12 @@ def get(self):
         for row in query_results:
             system_dict = row.System.__dict__
             host = {skey: system_dict[skey] for skey in system_columns}
-            host['recommendation_count'] = len(host['rule_hit_details'])
-            host['state'] = host['rule_hit_details'][0].get('key')
+            if host['rule_hit_details']:
+                host['recommendation_count'] = len(host['rule_hit_details'])
+                host['state'] = host['rule_hit_details'][0].get('key')
+            else:
+                host['recommendation_count'] = 0
```
kgaikwad, line 99: Will it be good to skip such systems which don't have any rule hits?

ros/api/v0/hosts.py
```diff
@@ -92,8 +92,12 @@ def get(self):
         for row in query_results:
             system_dict = row.System.__dict__
             host = {skey: system_dict[skey] for skey in system_columns}
-            host['recommendation_count'] = len(host['rule_hit_details'])
-            host['state'] = host['rule_hit_details'][0].get('key')
+            if host['rule_hit_details']:
```
apuntamb, line 95: @patilsuraj767 I do not quite understand that why do we get `host['rule_hit_details'] : None` even when there is a non-empty report that we have assigned to the `rule_hit_details` column under `systems` table and it does have the value in that record.

apuntamb: I guess you need to add `rule_hit_details` key [here](https://github.com/RedHatInsights/ros-backend/pull/52/files#diff-5b6c544a7a56034be87e719ba2ee43d4078f705de9c3049ee8dc5b7f8ee09382R32) for it to be available.

apuntamb (changes requested):
Added comments inline.

ros/api/v0/hosts.py
patilsuraj767: As we have two processors, it is possible that the inventory_event processor will process the report and will create the system entry under the system table and for some reason insights-engine-result processor might fail to populate the `rule_hit_details` in the system table. In this case host['rule_hit_details'] will return `None`.

ros/api/v0/hosts.py
patilsuraj767: Yes, we can do that.

ros/api/v0/hosts.py
```diff
@@ -92,8 +92,11 @@ def get(self):
...
+                host['recommendation_count'] = len(host['rule_hit_details'])
+                host['state'] = host['rule_hit_details'][0].get('key')
```
kgaikwad, line 97: @patilsuraj767, 
[Query] As we have one rule set which contains all rules, each rule hit will be having `key` value.
On frontend, each system row is having `state` column.  
Here you assigned `host['rule_hit_details'][0].get('key')` value to `state`, does that mean every `key` from rule hit contain same value?  OR need a additional logic to get `state`?

patilsuraj767: @kgaikwad  As discussed internally, for now we will be only having one rule hit.

kgaikwad (approved):
Codewise looks good to me :+1: 
Not tested.

kgaikwad:
I noticed one thing after merging these changes. 
Here you handled keyError inside `/systems` but no conditional handling in `/system/<host_id>` API. 
In case of systems without rule hits, while calculating values for columns - state & recommendation count, API request results into exception.

PR #53, opened 2021-03-17
RHIROS-85 Modify fields in Hosts and Details API responses

apuntamb:
Added response fields as per the mockup here: https://marvelapp.com/prototype/9408d9j/screen/74816603.
EDIT:

- [x] Add missing fields in **both** APIs such as `idling_time`, `io_wait`, `rule_hit_details`, `cloud_provider`, `report_date`, `instance_type`.
- [x] Pending existing fields such as `recommendation_count` and `state` to be displayed after #55 is merged.
- [x] Renamed `recommendation_count` to `number_of_recommendations`.
- [x] HostDetailsAPI will display host id with `inventory_id` as the key.

### DIFF
--- a/ros/api/v0/hosts.py
+++ b/ros/api/v0/hosts.py
@@ -30,10 +30,9 @@ class HostsApi(Resource):
         'state': fields.String,
         'display_performance_score': fields.Nested(display_performance_score_fields),
         'cloud_provider': fields.String,
-        'instance_type': fields.String
-        # TODO - idling_time, io_wait
-        # 'idling_time': fields.String,
-        # 'io_wait': fields.String
+        'instance_type': fields.String,
+        'idling_time': fields.String,
+        'io_wait': fields.String
     }
     meta_fields = {
         'count': fields.Integer,
@@ -113,6 +112,8 @@ class HostsApi(Resource):
                 continue
             host['account'] = row.RhAccount.account
             host['display_performance_score'] = row.PerformanceProfile.display_performance_score
+            host['idling_time'] = row.PerformanceProfile.idling_time
+            host['io_wait'] = row.PerformanceProfile.io_wait
             hosts.append(host)
 
         return build_paginated_system_list_response(
@@ -156,7 +157,6 @@ class HostsApi(Resource):
 class HostDetailsApi(Resource):
     profile_fields = {
         'host_id': fields.String(attribute='inventory_id'),
-        'performance_record': fields.String,
         'rating': fields.Integer,
         'report_date': fields.String,
         'instance_type': fields.String,
@@ -195,10 +195,8 @@ class HostDetailsApi(Resource):
             record['report_date'] = profile.report_date
             record['cloud_provider'] = system.cloud_provider
             record['instance_type'] = system.instance_type
-            record['idling_time'] = (profile.performance_record['kernel.all.cpu.idle']
-                                     / profile.performance_record['total_cpus']) * 100
-            # This metric has not been updated in our archive yet.
-            record['io_wait'] = profile.performance_record['kernel.all.cpu.wait.total'] * 100
+            record['idling_time'] = profile.idling_time
+            record['io_wait'] = profile.io_wait
         else:
             abort(404, message="System {} doesn't exist"
                   .format(host_id))
--- a/ros/api/v0/hosts.py
+++ b/ros/api/v0/hosts.py
@@ -15,11 +15,31 @@ DEFAULT_OFFSET = 0
 
 
 class HostsApi(Resource):
-
     display_performance_score_fields = {
         'cpu_score': fields.Integer,
         'memory_score': fields.Integer,
         'io_score': fields.Integer
+    }
+    link_key_fields = {
+        'kcs': fields.List,
+        'jira': fields.List
+    }
+    details_fields = {
+        'rhel': fields.String,
+        'type': fields.String,
+        'error_key': fields.String
+        # Are instance_type and cloud_provider needed here too?
+        # We have them below already.
+    }
+    rule_hit_details_fields = {
+        'key': fields.String,
+        'tags': fields.List,
+        'type': fields.String,
+        'links': fields.Nested(link_key_fields),
+        'details': fields.Nested(details_fields),
+        'rule_id': fields.String,
+        'component': fields.String,
+        'system_id': fields.String
     }
     hosts_fields = {
         'fqdn': fields.String,
@@ -33,7 +53,7 @@ class HostsApi(Resource):
         'instance_type': fields.String,
         'idling_time': fields.String,
         'io_wait': fields.String,
-        'rule_hit_details': fields.String
+        'rule_hit_details': fields.List(fields.Nested(rule_hit_details_fields))
     }
     meta_fields = {
         'count': fields.Integer,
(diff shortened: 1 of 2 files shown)


ros/api/v0/hosts.py
```diff
@@ -107,10 +107,12 @@ class HostDetailsApi(Resource):
     profile_fields = {
         'host_id': fields.String(attribute='inventory_id'),
         'performance_record': fields.String,
```
patilsuraj767, line 109: While testing I found `performance_record` is getting set to null. We can also remove `performance_record` from here.

patilsuraj767 (changes requested):
Thanks @apuntamb 
PR looks good, just add a small change request.

kgaikwad (changes requested):
One observation is that both `/systems` & `/systems/:inventory_ID` both required `idling_time` & `io_wait` fields. Don't you think both APIs need to be fixed?

I agree that this PR is about adding new fields to HostDetails API but in my opinion we should avoid repeat cycles for same things in these two APIs. Otherwise we will loose consistency between these two APIs.

ros/api/v0/hosts.py
apuntamb: done.

ros/api/v0/hosts.py
```diff
@@ -139,12 +141,16 @@ def get(self, host_id):
...
         if profile:
             record = {'inventory_id': host_id}
-            record['display_performance_score'] = profile.display_performance_score
             record['rating'] = rating_record.rating if rating_record else None
-            record['recommendation_count'] = len(system.rule_hit_details)
-            record['state'] = system.rule_hit_details[0].get('key')
+            record['report_date'] = profile.report_date
+            record['cloud_provider'] = system.cloud_provider
+            record['instance_type'] = system.instance_type
+            record['idling_time'] = (profile.performance_record['kernel.all.cpu.idle']
```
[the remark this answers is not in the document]
apuntamb: fixed at model level.

ros/lib/models.py
```diff
@@ -28,6 +28,15 @@ def display_performance_score(self):
...
         return display_performance_score
 
+    @property
+    def idling_time(self):
+        return (self.performance_record['kernel.all.cpu.idle']
+                / self.performance_record['total_cpus']) * 100
+
+    @property
+    def io_wait(self):
+        return self.performance_record['kernel.all.cpu.wait.total'] * 100
```
patilsuraj767, line 38: Here we are calculating idling_time and io_wait in `percentage` but in the mockups, they are shown in `mins`.

apuntamb: Good catch, fixed! :+1:

ros/api/v0/hosts.py
```diff
@@ -151,11 +152,12 @@ def build_sort_expression(self, order_how, order_method):
 class HostDetailsApi(Resource):
     profile_fields = {
         'host_id': fields.String(attribute='inventory_id'),
```
kgaikwad, line 154: I am getting key `inventory_id` in response of `/systems` API. Will it make sense to keep `inventory_id` as it is?

ros/api/v0/hosts.py
```diff
@@ -151,11 +152,12 @@ def build_sort_expression(self, order_how, order_method):
...
-        'performance_record': fields.String,
-        'display_performance_score': fields.String,
         'rating': fields.Integer,
-        'recommendation_count': fields.Integer,
```
kgaikwad, line 157: Any reason you not included keys - `display_performance_score, recommendation_count, state` in response?

ros/api/v0/hosts.py
```diff
@@ -184,12 +186,14 @@ def get(self, host_id):
 
         if profile:
             record = {'inventory_id': host_id}
-            record['display_performance_score'] = profile.display_performance_score
             record['rating'] = rating_record.rating if rating_record else None
-            record['recommendation_count'] = len(system.rule_hit_details)
-            record['state'] = system.rule_hit_details[0].get('key')
+            record['report_date'] = profile.report_date
```
kgaikwad, line 190: Could you please include rule hits details same as `/systems` API?

kgaikwad (changes requested):
@apuntamb, 
I found inconsistency between both APIs. `/systems/<host_id>` API can include more detailed information of a specific system not less. 
Few more inline comments added.

ros/api/v0/hosts.py
kgaikwad, line 157: Instead of `recommendation_count`, will it be good to rename this to `number_of_recommendations`?

ros/api/v0/hosts.py
apuntamb: You mean modifying the key in HostsApi to `host_id` like that in HostDetailsApi ?

ros/api/v0/hosts.py
apuntamb: Yeah, so I tried to keep the details keys consistent with this [mockup](https://marvelapp.com/prototype/9408d9j/screen/74816603). If we still want all other fields related to a host, we can include those. 
I just didn't find much use of that data as we are not showing the same on  details page UI. (If we are to consume this API for the details page).
Let me know if we still want those keys.

ros/api/v0/hosts.py
apuntamb: We do not have `rule_hit_details` **shown** in /systems response. It is just used to calculate recommendation count currently. And once we start storing the rec_count and state in the coulmn we might not need to include it in /systems call either.

ros/api/v0/hosts.py
kgaikwad: Sorry for confusion. I think instead of modifying HostsApi, don't rename it here in HostDetailsApi to `host_id`. 
What do you think?

apuntamb: Cool, makes sense.

ros/api/v0/hosts.py
kgaikwad: On system detail page, there is recommendations list below as per mockup. 
Please let me know how backend is going to send those details. As per this, frontend will consume data.

ros/api/v0/hosts.py
```diff
@@ -143,19 +146,24 @@ def build_sort_expression(self, order_how, order_method):
...
         abort(403, message="Unexpected sort method {}".format(order_method))
         return None
 
 
 class HostDetailsApi(Resource):
     profile_fields = {
-        'host_id': fields.String(attribute='inventory_id'),
-        'performance_record': fields.String,
+        'inventory_id': fields.String,
         'display_performance_score': fields.String,
```
kgaikwad, line 157: `display_performance_score` is of dict type & not of string type, right?

kgaikwad (changes requested):

ros/api/v0/hosts.py
```diff
@@ -15,25 +15,45 @@
...
         'display_performance_score': fields.Nested(display_performance_score_fields),
         'cloud_provider': fields.String,
-        'instance_type': fields.String
-        # TODO - idling_time, io_wait
-        # 'idling_time': fields.String,
-        # 'io_wait': fields.String
+        'instance_type': fields.String,
+        'idling_time': fields.String,
+        'io_wait': fields.String,
+        'rule_hit_details': fields.List(fields.Nested(rule_hit_details_fields))
```
kgaikwad, line 56: Just small correction - here you can skip field `rule_hit_details`.

ros/api/v0/hosts.py
```diff
@@ -186,10 +240,20 @@ def get(self, host_id):
...
             record['display_performance_score'] = profile.display_performance_score
             record['rating'] = rating_record.rating if rating_record else None
-            record['recommendation_count'] = len(system.rule_hit_details)
-            record['state'] = system.rule_hit_details[0].get('key')
+            record['report_date'] = profile.report_date
+            record['cloud_provider'] = system.cloud_provider
+            record['instance_type'] = system.instance_type
+            record['idling_time'] = profile.idling_time
+            record['io_wait'] = profile.io_wait
+            record['rule_hit_details'] = system.rule_hit_details
```
kgaikwad, line 248: Hello @apuntamb, @patilsuraj767 
As discussed, we might need a separate API for recommendations. In that case, will it be good to skip field `rule_hits_detail` here as well?

ros/lib/models.py
```diff
@@ -28,6 +28,17 @@ def display_performance_score(self):
 
         return display_performance_score
 
+    @property
+    def idling_time(self):
+        idling_percent = (self.performance_record['kernel.all.cpu.idle']
+                          / self.performance_record['total_cpus']) * 100
+        # idling % of 24 hours and converting hours into mins
+        return 24 * (idling_percent / 100) * 60
```
kgaikwad, line 36: What is unit of `self.performance_record['kernel.all.cpu.idle']`?

apuntamb: As discussed, modifying it to just %

kgaikwad (approved):
LGTM :+1:

kgaikwad:
Thanks @apuntamb! Could you please review & merge respective frontend change - RedHatInsights/ros-frontend/pull/25?

PR #54, opened 2021-03-22
RHIROS-81 Implement basic filtering for systems

apuntamb:
Currently, it works fine for API. For example if we hit: `curl -v -H "Content-Type: application/json"  -H "x-rh-identity: ${b64_identity}" localhost:8080/api/ros/v0/systems?display_name=ap-172-31-11-67.ap-south-1.compute.internal | python -m json.tool`

### DIFF
--- a/ros/api/v0/hosts.py
+++ b/ros/api/v0/hosts.py
@@ -61,6 +61,9 @@ class HostsApi(Resource):
         ).strip().lower()
         order_how = (request.args.get('order_how') or 'asc').strip().lower()
 
+        # if query string is passed with display_name
+        filter_display_name = request.args.get('display_name')
+
         ident = identity(request)['identity']
         # Note that When using LIMIT, it is important to use an ORDER BY clause
         # that constrains the result rows into a unique order.
@@ -106,6 +109,11 @@ class HostsApi(Resource):
             host['account'] = row.RhAccount.account
             host['display_performance_score'] = row.PerformanceProfile.display_performance_score
             hosts.append(host)
+            # for filtering get the request param and check
+            if filter_display_name and host['display_name'] == filter_display_name:
+                hosts = []
+                hosts.append(host)
+                break
 
         return build_paginated_system_list_response(
             limit, offset, hosts, count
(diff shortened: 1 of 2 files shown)


ros/api/v0/hosts.py
```diff
@@ -97,6 +99,11 @@ def get(self):
             host['account'] = row.RhAccount.account
             host['display_performance_score'] = row.PerformanceProfile.display_performance_score
             hosts.append(host)
+            # for filtering get the request param and check
+            if filter_display_name and host['display_name'] == filter_display_name:
```
kgaikwad, line 103: Will it be good to check this condition using `where` clause in query itself?

kgaikwad (changes requested):
@apuntamb, could you please confirm how to send `display_name` filter in request?
Pull-request description is not matching with openapi spec.

apuntamb:
> @apuntamb, could you please confirm how to send `display_name` filter in request?
> Pull-request description is not matching with openapi spec.

done

ros/api/v0/hosts.py
```diff
@@ -68,13 +69,18 @@ def get(self):
         # Refer - https://www.postgresql.org/docs/13/queries-limit.html
 
         account_query = db.session.query(RhAccount.id).filter(RhAccount.account == ident['account_number']).subquery()
-        system_query = db.session.query(System.id)\
-            .filter(System.account_id.in_(account_query)).subquery()
+        if filter_display_name:
+            system_query = db.session.query(System.id)\
+                .filter(System.account_id.in_(account_query))\
+                .filter(System.display_name == filter_display_name)
```
kgaikwad, line 75: With this change, end user needs to pass whole `display_name` to get matching systems. Is it correct?
OR need a full-text search within this field?

apuntamb: Yes, currently end user needs to pass the whole `display_name` to get the matched system.

kgaikwad: @apuntamb,
It would be good to have `full text + lowercase`. 

For lowercase, you can check on `ilike` - https://docs.sqlalchemy.org/en/13/orm/internals.html#sqlalchemy.orm.PropComparator.ilike

For full search, you can make use of `%` escape character. Example - "%foobar%"

apuntamb: done.

kgaikwad (approved):
@apuntamb, thanks for updates.
Codewise looks good to me now :+1:

PR #48, opened 2021-03-10
Get state and recommendations count from insights rule_hit(DB).

patilsuraj767:
This PR should be only merged after insights-engine(advisor) integration PR - https://github.com/RedHatInsights/ros-backend/pull/46

PR #55, opened 2021-03-25
RHIROS-86 Refactor storing and using recommendations and state fields

apuntamb:
Handles overall refactoring of recommendations and state fields.
- [x] Add separate columns for state & number_of_recommendations.
- [x] While report processing add and update both values along with rule hits.
- [x] When no rule hits for ROS rules, processor should delete the system.
- [x] Sort systems based on recommendations and state columns.
- [x] verify renaming all `recommendation_count`s to `number_of_recommendations` in the end before merging.


ros/api/v0/hosts.py
```diff
@@ -139,8 +139,15 @@ def build_sort_expression(self, order_how, order_method):
...
                 asc(PerformanceProfile.system_id),)
-        # FIXME: no ordering as of now for columns:
-        # state, recommendation_count
+
+        if order_method == 'recommendation_count':
+            return (sort_order(System.rule_hit_details),
+                    asc(PerformanceProfile.system_id),)
+
+        if order_method == 'state':
+            return (sort_order(System.rule_hit_details[0]['key']),
```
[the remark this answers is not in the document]
apuntamb: Sure, I'll handle if no `rule_hit_details` are present. 
Well I'm not sure but, as per [#48](https://github.com/RedHatInsights/ros-backend/pull/48#discussion_r594150821), the rule_hit_details change inside systems table was merged there.

ros/api/v0/hosts.py
```diff
@@ -139,8 +139,15 @@ def build_sort_expression(self, order_how, order_method):
                 sort_order(PerformanceProfile.performance_score[
                     order_method].astext.cast(Integer)),
                 asc(PerformanceProfile.system_id),)
-        # FIXME: no ordering as of now for columns:
-        # state, recommendation_count
+
+        if order_method == 'recommendation_count':
+            return (sort_order(System.rule_hit_details),
```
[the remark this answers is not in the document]
apuntamb: recommedation_count is gathered from the length of the array of all rule hits: [#48](https://github.com/RedHatInsights/ros-backend/pull/48/files#diff-5b6c544a7a56034be87e719ba2ee43d4078f705de9c3049ee8dc5b7f8ee09382R95)

ros/api/v0/hosts.py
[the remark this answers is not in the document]
patilsuraj767: @kgaikwad I did not get your point. What difference has a separate table for rule_hit_details have made? 
If having a separate `state` column in the system table is the suggestion(which will make sorting easy) then I am fine with that implementation.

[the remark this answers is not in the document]
kgaikwad: > If having a separate `state` column in the system table is the suggestion(which will make sorting easy) then I am fine with that implementation.

@patilsuraj767 - Yes, it will become easy. This was just a thought. 

As here sorting for `state` column is based on nested key of rule_hit_details, when no rule_hit_details captured for any system(s), will it simply return all records or give any warning/exception?
^^ this is what I wanted to confirm in my 1st query.
If it is going to return all records then @apuntamb you can skip this comment.

ros/api/v0/hosts.py
```diff
@@ -147,8 +147,15 @@ def build_sort_expression(self, order_how, order_method):
                 sort_order(PerformanceProfile.performance_score[
                     order_method].astext.cast(Integer)),
                 asc(PerformanceProfile.system_id),)
-        # FIXME: no ordering as of now for columns:
-        # state, recommendation_count
+
+        if order_method == 'number_of_recommendations':
+            return (sort_order(System.rule_hit_details),
```
apuntamb, line 152: TODO:
```suggestion
            return (sort_order(System.number_of_recommendations),
```

ros/api/v0/hosts.py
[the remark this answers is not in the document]
apuntamb: Created `state` and `number_of_recommendations` columns in Systems table for easy sorting and searching (for later time).

ros/api/v0/hosts.py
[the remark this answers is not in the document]
apuntamb: now we have a separate column for the count.

ros/lib/utils.py
```diff
@@ -31,6 +31,21 @@ def get_or_create(session, model, keys, **kwargs):
     return instance
 
 
+def delete_record(session, model, keys, **kwargs):
```
kgaikwad, line 34:
```suggestion
def delete_record(session, model, **kwargs):
```
I don't think this method requires keys argument. can't you simply iterate on `kwargs`

kgaikwad, line 34: will it be good to rename `kwargs` to something specific? OR add details to method docstring?

ros/processor/insights_engine_result_consumer.py
```diff
@@ -66,8 +66,13 @@ def handle_msg(self, msg):
                 for report in reports:
                     if 'cloud_instance_ros_evaluation' in report["rule_id"]:
                         ros_reports.append(report)
-
-                self.process_report(host, ros_reports)
+                if len(ros_reports) == 0:
+                    with app.app_context():
+                        deleted_id = delete_record(db.session, System, 'inventory_id',
+                                                   inventory_id=host['id'])
+                        LOG.info("Deleted system with inventory_id: %s via engine-result", deleted_id)
```
kgaikwad, line 73:
```suggestion
                        inventory_id = delete_record(db.session, System, 'inventory_id',
                                                   inventory_id=host['id'])
                        LOG.info("Deleted system with inventory_id: %s via engine-result", inventory_id)
```


Any impact after removing a whole record from System table? I guess it will recreate System record next time.

ros/lib/utils.py
```diff
@@ -31,6 +31,21 @@ def get_or_create(session, model, keys, **kwargs):
...
+    if not keys:
+        keys = kwargs.keys()
+    if isinstance(keys, str):
+        keys = [keys]
+    if not isinstance(keys, list):
+        raise TypeError('keys argument must be a string')
+    instance = session.query(model).filter_by(**{k: kwargs[k] for k in keys}).first()
+    deleted = instance.inventory_id
```
kgaikwad, line 42: Here you assumed that every instance of any model have `inventory_id` column. Is this correct?

kgaikwad (changes requested):

ros/processor/insights_engine_result_consumer.py
apuntamb: Yeah, no such impact. It would create a new record next time.

ros/lib/utils.py
apuntamb: Yeah. Are there any cases where we won't have inventory_id for a system?

ros/api/v0/hosts.py
```diff
@@ -104,12 +104,6 @@ def get(self):
         for row in query_results:
             system_dict = row.System.__dict__
             host = {skey: system_dict[skey] for skey in system_columns}
-            if host['rule_hit_details']:
-                # FIXME after Ticket-86 is resolved
```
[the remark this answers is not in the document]
apuntamb: Not expected, I missed that. Corrected now. :+1:

ros/lib/utils.py
apuntamb: updated.

apuntamb: renamed to `system_fields`.

ros/lib/utils.py
```diff
@@ -31,6 +31,16 @@ def get_or_create(session, model, keys, **kwargs):
...
 
 
+def delete_record(session, model, **system_fields):
+    keys = list(system_fields.keys())
+    instance = session.query(model).filter_by(**{k: system_fields[k] for k in keys}).first()
+    inventory_id = instance.inventory_id
+    if instance:
+        session.delete(instance)
+        session.commit()
+    return inventory_id
```
patilsuraj767, line 41: Lets make `delete_record` function more generic and not specific to systems table. 

```suggestion
def delete_record(session, model, **kwargs):
    keys = list(kwargs.keys())
    instance = session.query(model).filter_by(**{k: kwargs[k] for k in keys}).first()
    if instance:
        session.delete(instance)
        session.commit()
    return instance
```

ros/processor/insights_engine_result_consumer.py
```diff
@@ -66,8 +66,12 @@ def handle_msg(self, msg):
                 for report in reports:
                     if 'cloud_instance_ros_evaluation' in report["rule_id"]:
                         ros_reports.append(report)
-
-                self.process_report(host, ros_reports)
+                if len(ros_reports) == 0:
+                    with app.app_context():
+                        deleted_inventory_id = delete_record(db.session, System, inventory_id=host['id'])
+                        LOG.info("Deleted system with inventory_id: %s via engine-result", deleted_inventory_id)
```
patilsuraj767, line 72:
```suggestion
                        system = delete_record(db.session, System, inventory_id=host['id'])
                        LOG.info("Deleted system with inventory_id: %s via engine-result", system.inventory_id)
```

patilsuraj767 (changes requested):

ros/lib/utils.py
apuntamb: As per Suraj's suggestion [here](https://github.com/RedHatInsights/ros-backend/pull/55/#discussion_r610355151), I'll keep it to kwargs but, add details to the method.

ros/lib/utils.py
apuntamb: Now, that we are returning the instance itself, we need not worry for the presence/absence of `inventory_id`.

patilsuraj767 (approved):

kgaikwad (approved):
@apuntamb @patilsuraj767 -
For now, changes look good.
But I am still have few concerns around system record deletion change instead of clearing fields. So marking one of my comment as `Unresolved`

ros/processor/insights_engine_result_consumer.py
kgaikwad: Marked this comment as `Unresolved` because having concern around system record deletion change instead of clearing fields.  Currently ROS DB is the only place to get info about for how many systems ROS is enabled.

patilsuraj767:
Thanks @apuntamb  :+1: 
Merged

PR #56, opened 2021-03-26
Inital setup for unit test in ROS

patilsuraj767:


tests/fixtures/db_fixtures.py
```diff
@@ -0,0 +1,67 @@
...
+        display_name='ip-172-31-11-67.ap-south-1.compute.internal',
+        fqdn='ip-172-31-11-67.ap-south-1.compute.internal',
+        cloud_provider='aws',
+        instance_type='t2.micro',
+        rule_hit_details='{"rule_id": "cloud_instance_ros_evaluation|CONSUMPTION_MODEL","component":'
+                         '"telemetry.rules.plugins.ros.cloud_instance_ros_evaluation.report_consumption_model",'
+                         '"type": "rule","key": "CONSUMPTION_MODEL","details": {"rhel": "8.3","cloud_provider": "aws",'
+                         '"instance_type": "t2.micro","type": "rule","error_key": "CONSUMPTION_MODEL"},"tags": [],'
+                         '"links": {"jira": ["https://issues.redhat.com/browse/CEECBA-5092"],"kcs": ["T.B.D"]},'
+                         '"system_id": "677fb960-e164-48a4-929f-59e2d917b444"}')
```
apuntamb, line 58: Can you add `state` and `number_of_recommendations` column here?

tests/fixtures/db_fixtures.py
```diff
@@ -0,0 +1,67 @@
+import os
+import pytest
+from ros.lib.app import app, db
+from sqlalchemy_utils import database_exists, create_database, drop_database
+from ros.lib.models import RhAccount, System
```
apuntamb, line 5: Just a small  query - don't we need a test for models?

patilsuraj767: No, There is no need to test the models.

tests/fixtures/db_fixtures.py
patilsuraj767: Ok, I will add those columns.

apuntamb (approved):
Works fine!
LGTM :+1:

apuntamb:
Merged, thanks @patilsuraj767 :+1: :100:

PR #58, opened 2021-03-30
RHIROS-101 - Format state column as per mockup.

patilsuraj767:


ros/api/v0/hosts.py
```diff
@@ -187,7 +191,8 @@ def get(self, host_id):
             record['display_performance_score'] = profile.display_performance_score
             record['rating'] = rating_record.rating if rating_record else None
             record['recommendation_count'] = len(system.rule_hit_details)
-            record['state'] = system.rule_hit_details[0].get('key')
+            rule_error_key = system.rule_hit_details[0].get('key')
+            record['state'] = SYSTEM_STATES[rule_error_key]
```
apuntamb, line 195: As per this [mockup](https://marvelapp.com/prototype/9408d9j/screen/74816603), do we need the state field on the details page?

apuntamb:
Are we not adding `state` column to the systems table here?

ros/api/v0/hosts.py
patilsuraj767: Yeah, as per mockups I think we do not need it in `HostDetailsApi`. I will remove the `state` from here.

ros/api/v0/hosts.py
```diff
@@ -13,6 +13,9 @@
 DEFAULT_HOSTS_PER_REP = 10
 DEFAULT_OFFSET = 0
 
+SYSTEM_STATES = {"INSTANCE_OVERSIZED": "Oversized", "INSTANCE_UNDERSIZED": "Undersized",
+                 "CONSUMPTION_MODEL": "Idling", "STORAGE_RIGHTSIZING": "Storage rightsizing"}
```
apuntamb, line 17: If we are to store the `state` in `systems` table, shouldn't this be defined in the processor directly?

patilsuraj767: Currently, the state is not stored in a separate column, we are storing the JSON object in the rule_hit_details column as is.
My thought was not to store modified JSON data(rule hit) in DB but to modify it at the time of fetching through API.

apuntamb:
Closing in favor of #55.

PR #60, opened 2021-03-31
RHIROS-102 Performance score should always be >= 1

patilsuraj767:


apuntamb (approved):

apuntamb:
Merged, thanks for the quick fix @patilsuraj767 !

PR #70, opened 2021-04-28
RHIROS-130 Add support for Azure in ROS

apuntamb:
Attaching screnshot for reference:

azure_vm


patilsuraj767:
Tested the PR, LGTM :+1:  
@apuntamb Can you please rebase this PR, I would like to run the unit test on this PR just to confirm that it is not breaking anything.

apuntamb:
> Tested the PR, LGTM +1
> @apuntamb Can you please rebase this PR, I would like to run the unit test on this PR just to confirm that it is not breaking anything.

Thanks for reviewing the PR @patilsuraj767 ! I've rebased the PR but I'm not able to run the tests locally as somehow ros-processor in my local isn't running. (Might be irrelevant to the tests or this PR). 
But, I would suggest we should have tests running in our build i.e. we should add `pipenv run pytest --cov=ros tests` in `.travis.yml `to check if any existing tests fail because of a new change.

patilsuraj767:
@apuntamb As discussed ros-processor problem should be specific to your local dev environment and yes, we will add pytest in travis job. 
As I have tested this PR locally and it is working, I am going ahead and merging it. 
Thanks @apuntamb :+1:

PR #73, opened 2021-05-11
Fixes #145 Add core_collection check in ansible playbook

patilsuraj767:


apuntamb:
@kgaikwad This PR only checks if insights-client is installed or not and if core_collect is enables or not on the system. This means users should install and enable these on their own? Is that the expected behaviour?

kgaikwad:
@apuntamb, could you please do rebase & create new pull-request?
Close this one on behalf of new PR.

apuntamb:
Closing this in favor of #83.
Thanks @patilsuraj767!

PR #74, opened 2021-05-18
API to list historical performance data

patilsuraj767:
API output looks like below - 

```
{
    "meta": {
        "count": 3,
        "limit": 10,
        "offset": 0
    },
    "links": {
        "first": "/api/ros/v0/systems/02c4bc40-e507-4b35-a321-732bfe74ba91/history?limit=10&offset=0",
        "last": "/api/ros/v0/systems/02c4bc40-e507-4b35-a321-732bfe74ba91/history?limit=10&offset=0",
        "next": null,
        "previous": null
    },
    "data": {
        "inventory_id": "02c4bc40-e507-4b35-a321-732bfe74ba91",
        "performance_history": [
            {
                "cpu_score": 1,
                "memory_score": 3,
                "io_score": 1,
                "report_date": "2021-05-18"
            },
            {
                "cpu_score": 1,
                "memory_score": 3,
                "io_score": 1,
                "report_date": "2021-05-17"
            },
            {
                "cpu_score": 5,
                "memory_score": 3,
                "io_score": 5,
                "report_date": "2021-05-16"
            }
        ]
    }
}
```

### DIFF
--- a/ros/api/v0/hosts.py
+++ b/ros/api/v0/hosts.py
@@ -222,3 +222,71 @@ class HostDetailsApi(Resource):
                   .format(host_id))
 
         return record
+
+
+class HostHistoryApi(Resource):
+    display_performance_score_fields = {
+        'cpu_score': fields.Integer,
+        'memory_score': fields.Integer,
+        'io_score': fields.Integer,
+        'report_date':  fields.String
+    }
+
+    data_fields = {
+        'inventory_id': fields.String,
+        'performance_history': fields.List(fields.Nested(display_performance_score_fields))
+    }
+    meta_fields = {
+        'count': fields.Integer,
+        'limit': fields.Integer,
+        'offset': fields.Integer
+    }
+    links_fields = {
+        'first': fields.String,
+        'last': fields.String,
+        'next': fields.String,
+        'previous': fields.String
+    }
+    history_fields = {
+        'meta': fields.Nested(meta_fields),
+        'links': fields.Nested(links_fields),
+        'data': fields.Nested(data_fields)
+    }
+
+    @marshal_with(history_fields)
+    def get(self, host_id):
+        limit = int(request.args.get('limit') or DEFAULT_HOSTS_PER_REP)
+        offset = int(request.args.get('offset') or DEFAULT_OFFSET)
+        if not is_valid_uuid(host_id):
+            abort(404, message='Invalid host_id, Id should be in form of UUID4')
+
+        ident = identity(request)['identity']
+
+        account_query = db.session.query(RhAccount.id).filter(RhAccount.account == ident['account_number']).subquery()
+        system_query = db.session.query(System.id) \
+            .filter(System.account_id.in_(account_query)).filter(System.inventory_id == host_id).subquery()
+
+        query = PerformanceProfile.query.filter(
+            PerformanceProfile.system_id.in_(system_query)
+        ).order_by(PerformanceProfile.report_date.desc())
+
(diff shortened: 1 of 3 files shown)


kgaikwad (changes requested):

ros/api/v0/hosts.py
```diff
@@ -222,3 +222,71 @@ def get(self, host_id):
...
+    links_fields = {
+        'first': fields.String,
+        'last': fields.String,
+        'next': fields.String,
+        'previous': fields.String
+    }
+    history_fields = {
+        'meta': fields.Nested(meta_fields),
+        'links': fields.Nested(links_fields),
+        'data': fields.Nested(data_fields)
```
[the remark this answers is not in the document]
patilsuraj767: We cannot do this in case of API which supports pagination, because `build_paginated_system_list_response` only accept one parameter to construct the output response.

ros/api/v0/hosts.py
```diff
@@ -222,3 +222,71 @@ def get(self, host_id):
...
+
+            record = {'inventory_id': host_id}
+            record['performance_history'] = performance_history
+        else:
+            abort(404, message="System {} doesn't exist"
+                  .format(host_id))
+
+        return build_paginated_system_list_response(
+            limit, offset, record, count
+        )
```
kgaikwad, line 292: @patilsuraj767, with reference to previous comment -

> We cannot do this in case of API which supports pagination, because build_paginated_system_list_response only accept one parameter to construct the output response.

From this API, it will return historical data of a specific system using `inventory_id` and `build_paginated_system_list_response` method returns constructed JSON object so will it be good idea to add `inventory_id` key into that JSON object?


```suggestion
        paginated_response =  build_paginated_system_list_response(
            limit, offset, performance_history, count
        )
        paginated_response['inventory_id'] = host_id
```

ros/api/v0/hosts.py
```diff
@@ -222,3 +222,71 @@ def get(self, host_id):
...
+            for profile in query_results:
+                performance_record = profile.display_performance_score
+                performance_record['report_date'] = profile.report_date
+                performance_history.append(performance_record)
+
+            record = {'inventory_id': host_id}
+            record['performance_history'] = performance_history
+        else:
+            abort(404, message="System {} doesn't exist"
+                  .format(host_id))
```
kgaikwad, line 288:
```suggestion
        if not query_results:
            abort(404, message="System {} doesn't exist"
                  .format(host_id))
        performance_history = []
        for profile in query_results:
               performance_record = profile.display_performance_score
               performance_record['report_date'] = profile.report_date
               performance_history.append(performance_record)
         
         
       
```

ros/api/v0/hosts.py
```diff
@@ -274,19 +274,20 @@ def get(self, host_id):
...
                   .format(host_id))
 
-        return build_paginated_system_list_response(
+        performance_history = []
+        for profile in query_results:
+            performance_record = profile.display_performance_score
+            performance_record['report_date'] = profile.report_date
+            performance_history.append(performance_record)
+
+        record = {'performance_history': performance_history}
```
kgaikwad, line 287: You could directly pass `performance_history` in data and no need to declare `record`, right?

kgaikwad (changes requested):
@patilsuraj767, PR looks good to me :+1:  
The only remaining part is to have openapi changes for this new API.

kgaikwad (approved):
LGTM :+1:

PR #80, opened 2021-06-14
RHIROS-173 Add rbac support

apuntamb:
Adding @beav for review.


ros/lib/config.py
```diff
@@ -25,3 +25,8 @@
 APP_NAME = os.getenv("APP_NAME", "ros")
 
 INSIGHTS_EXTRACT_LOGLEVEL = os.getenv("INSIGHTS_EXTRACT_LOGLEVEL", "ERROR")
+
+# TODO
+RBAC_HOST = os.getenv("RBAC_HOST", "localhost")
+RBAC_PORT = os.getenv("RBAC_PORT")
+RBAC_SVC_URL = f"http://{RBAC_HOST}:{RBAC_PORT}/"
```
beav, line 32: does this URL become invalid if RBAC_PORT is not set?

apuntamb: missed adding the port, it will default to 8114 now.

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,138 @@
...
+    logger.debug("fetched %s" % url)
+    _validate_service_response(response, logger, auth_header)
+    return response.json()
+
+
+def _validate_service_response(response, logger, auth_header):
+    """
+    Raise an exception if the response was not what we expected.
+    """
+    print(response.status_code)
```
apuntamb, line 34: leftover

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,138 @@
...
+    if response.status_code == requests.codes.not_found:
+        abort(
+            response.status_code, message="%s error received from service: %s" % response.text
```
kgaikwad, line 37: Will this line execute without an error?

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,138 @@
...
+            else:
+                logger.info("no identity or no key")
+        abort(
+            response.status_code, message=response.text
+        )
+
+    if response.status_code != requests.codes.ok:
+        abort(
+            response.status_code, message="Error received from backend service"
+        )
```
kgaikwad, line 57: Will it be possible to keep consistency between message values? 
Definitely, `response.text` value will be different for each exception but on line +41, you have added substring `error received from service: `, line +51 without this substring and here it includes somewhat same substring.

ros/lib/rbac_interface.py
```diff
@@ -104,18 +105,23 @@ def ensure_has_permission(**kwargs):
...
+            # if wrong permissions
+            abort(
                 HTTPStatus.FORBIDDEN,
-                message="user does not have access to %s" % kwargs["permissions"],
+                message='user does not have access to %s' % kwargs["permissions"]
             )
-        except RBACDenied:
-            raise HTTPError(
+        # if something wrong inside try e.g. perms not of type array
+        except Exception:
```
kgaikwad, line 115: Here, you have added  `Exception` that means this block will get executed when there is an exception. Any specific reason that you have passed status code `HTTPStatus.FORBIDDEN` to abort helper method for any kind of exception?

kgaikwad (changes requested):
@apuntamb,  adding few inline comments before testing the changes.

ros/lib/rbac_interface.py
beav: What does the `abort` call do?

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+RBAC_SVC_ENDPOINT = "/api/rbac/v1/access/?application=%s"
+AUTH_HEADER_NAME = "X-RH-IDENTITY"
+VALID_HTTP_VERBS = ["get", "options", "head", "post", "put", "patch", "delete"]
+
+
+def fetch_url(url, auth_header, logger, method="get"):
+    """
+    Helper to make a single request *** Add to this comment once functionality done.
```
beav, line 15: looks like a TODO item snuck in here. IMO either update the method doc block or just remove the "add to this comment" note if no additional info is needed.

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,138 @@
...
+    Raise an exception if the response was not what we expected.
+    """
+    print(response.status_code)
+    if response.status_code == requests.codes.not_found:
+        abort(
+            response.status_code, message="%s error received from service: %s" % response.text
+        )
+
+    if response.status_code in [requests.codes.forbidden, requests.codes.unauthorized]:
+        logger.info(
```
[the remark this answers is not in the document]
beav: If it gets an exception that's not a 403 or 401, it should bail out and log an error level message (i think)

beav:
@apuntamb I added a couple of comments but overall I think @kgaikwad has everything covered. Once the comments are addressed the PR looks good to me :+1:

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,138 @@
...
+def fetch_url(url, auth_header, logger, method="get"):
+    """
+    Helper to make a single request *** Add to this comment once functionality done.
+    """
+    if method not in VALID_HTTP_VERBS:
+        abort(
+            HTTPStatus.METHOD_NOT_ALLOWED, message="'%s' is not valid HTTP method." % method
+        )
+
+    logger.debug("Fetching %s" % url)
```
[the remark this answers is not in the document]
apuntamb: I think it won't be a problem even if we remove it before and after the request. Removing.

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,138 @@
...
+
+    response = requests.request(method, url, headers=auth_header)
+    logger.debug("fetched %s" % url)
```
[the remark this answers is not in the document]
apuntamb: ditto, removing.

ros/lib/rbac_interface.py
apuntamb: done.

ros/lib/rbac_interface.py
apuntamb: done

ros/lib/rbac_interface.py
apuntamb: @beav The abort helper will wrap an HTTP Exception with the status and message provided in the method call. https://flask-restful.readthedocs.io/en/latest/api.html#flask_restful.abort
@kgaikwad Do you mean will this line execute if there is no error? Also, I have made changes in the structure of if-else here.

ros/lib/rbac_interface.py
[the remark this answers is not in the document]
apuntamb: added log infos and in case of exceptions other than 401 or 403, logging errors and aborting the request.

ros/lib/rbac_interface.py
apuntamb: done.

ros/lib/rbac_interface.py
apuntamb: As you can see now, I have replaced the `Exception` with a custom exception called `RBACDenied`. And at this place, the exception raised should be related to access forbidden and hence, I added Forbidden.

ros/lib/rbac_interface.py
kgaikwad: > @kgaikwad Do you mean will this line execute if there is no error? Also, I have made changes in the structure of if-else here.

@apuntamb, I meant when there is 404 status code, will this below line run?
 `abort(response.status_code, message="%s error received from service: %s" % response.text)` 
 
You already modified the message so it won't be an issue now. 
But for your understanding, could you please run below statement into python shell? 
`"%s error received from service: %s" % response.text`

As per my understanding it should error out as you have passed a single argument to message string.

ros/lib/rbac_denied_exception.py
```diff
@@ -0,0 +1,8 @@
+class RBACDenied(Exception):
+    def __init__(self, message):
+        """
+        Raise this exception if the inventory service reports that you do not
```
kgaikwad, line 4: @apuntamb - do we need this exception class? 
The reason behind asking this question is doc string says that raise it if inventory service reports no permission to access. IIRC, ROS backend don't request any inventory API.

@beav, any thoughts?

ros/lib/rbac_interface.py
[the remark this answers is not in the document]
kgaikwad: Yeah, +1 to error log.

ros/lib/rbac_denied_exception.py
kgaikwad: In addition, here also I didn't see @apuntamb raising this exception anywhere.

ros/lib/rbac_interface.py
apuntamb: Ahh okk, I got your concern now. Yes there were two formatters `%s` added and only one value passed, so yes it would error out. I had made certain changes post testing so, mistakenly forgot to correct the statement. :)
I have corrected the same in latest commit.

ros/lib/rbac_denied_exception.py
apuntamb: @kgaikwad The motive is to raise a specific exception only for this `try block` [here](https://github.com/RedHatInsights/ros-backend/pull/80/files#diff-55bc290eed399bee5e1cb3fbc4494c7c218710af4a7adbc2c8cf3c5e92ad6e21R96-R112). I do not want to catch it in a generic `Exception` class because in that case any exception even though with its specific abort statement will be caught here. If you see I'm raising this RBACDenied on line-112.

Should I not use a try-except instead and abort the request where necessary? In that case I suppose I need not create a custom class. What do you think?

kgaikwad: > If you see I'm raising this RBACDenied on line-112.

From [line-112]( https://github.com/RedHatInsights/ros-backend/pull/80/files#diff-55bc290eed399bee5e1cb3fbc4494c7c218710af4a7adbc2c8cf3c5e92ad6e21R112), you have used `except` block.

The except block lets you handle the error. To raise an exception, one needs to use the `raise` keyword.

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,134 @@
...
+                if p in kwargs["permissions"]:
+                    return  # allow
+
+            # if wrong permissions
+            abort(
+                HTTPStatus.FORBIDDEN,
+                message='user does not have correct permissions to access the service'
+            )
+        # if something wrong inside `try`
+        except requests.exceptions.HTTPError:
```
kgaikwad, line 111: Except this block, everything else looks good to me.

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,134 @@
...
+            abort(
+                HTTPStatus.FORBIDDEN,
+                message="request to retrieve permissions from RBAC was forbidden"
+            )
+    else:
+        # if no auth_key
+        abort(
+            HTTPStatus.BAD_REQUEST, message="identity not found on request"
```
gnurag, line 119: Perhaps change the message to _identity not found **in** request_.

ros/lib/rbac_interface.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+        if not perms:
+            # if wrong permissions
+            abort(
+                HTTPStatus.FORBIDDEN,
+                message='User does not have correct permissions to access the service'
+            )
+        for p in perms:
+            if p in kwargs["permissions"]:
+                return  # allow
```
kgaikwad, line 108: @apuntamb  -
If there is matching permission, it will return correctly at line +180.  But what if there is no matching result? In that case, also it should raise forbidden exception.  Any thoughts?

kgaikwad (approved):
LGTM :+1: 
Thank you @apuntamb  for addressing all comments.

PR #81, opened 2021-06-14
RHIROS-180 fixes 404 error shown when no recommendations

apuntamb:


ros/api/v0/recommendations.py
```diff
@@ -63,5 +63,4 @@ def get(self, host_id):
                   'meta': {'count': len(recommendations_list)}
             }
         else:
-            abort(404, message="host with id {} doesn't have any recommendation"
-                  .format(host_id))
+            recommendations_list = []
```
kgaikwad, line 66: @apuntamb - what is the expected return value in this case?

apuntamb: return value should be common for both cases. added a commit for the same.

ros/api/v0/recommendations.py
```diff
@@ -56,11 +56,11 @@ def get(self, host_id):
                     for skey in rules_columns:
                         recommendation[skey] = eval("f'{}'".format(rule_dict[skey]))
                     recommendations_list.append(recommendation)
+        else:
+            recommendations_list = []
```
kgaikwad, line 60: No need of `else` block as `recommendations_list` already initialized to empty array.

kgaikwad (approved):
Thanks @apuntamb! LGTM :+1:

PR #84, opened 2021-06-23
Fixes #141 - Garbage collector to delete performance data older than 14days

kgaikwad:


kgaikwad:
Tested the changes locally and removed leftover. Merging it

PR #85, opened 2021-06-24
RHIROS-171 Ansible tasks to support ROS setup on RHEl7

apuntamb:
rhel-7-playbook

### DIFF
--- a/ansible-playbooks/ros_install_and_set_up.yml
+++ b/ansible-playbooks/ros_install_and_set_up.yml
@@ -47,20 +47,11 @@
         - ansible_architecture == "aarch64"
         - ansible_distribution == "RedHat"
         - ansible_distribution_major_version == "8"
-    - name: Enable required repository for RHEL7 x86_64 arch
+    - name: Enable required repository for RHEL7
       rhsm_repository:
         name: rhel-7-server-rpms
         state: enabled
       when:
-        - ansible_architecture == "x86_64"
-        - ansible_distribution == "RedHat"
-        - ansible_distribution_major_version == "7"
-    - name: Enable required repository for RHEL7 ARM arch
-      rhsm_repository:
-        name: rhel-7-server-rpms
-        state: enabled
-      when:
-        - ansible_architecture == "aarch64"
         - ansible_distribution == "RedHat"
         - ansible_distribution_major_version == "7"
     - name: Install pcp packages for ROS service
--- a/ansible-playbooks/ros_install_and_set_up.yml
+++ b/ansible-playbooks/ros_install_and_set_up.yml
@@ -28,17 +28,41 @@
     - fail:
         msg: insights-client core collect is disabled.
       when: core_collect.msg == 'line added'
+    - debug: var=ansible_distribution
+    - debug: var=ansible_distribution_major_version
     - debug: var=ansible_architecture
-    - name: Enable required repository for x86_64 arch
+    - name: Enable required repository for RHEL8 x86_64 arch
       rhsm_repository:
         name: rhel-8-for-x86_64-baseos-rpms,rhel-8-for-x86_64-appstream-rpms
         state: enabled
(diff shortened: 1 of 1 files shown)


ansible-playbooks/ros_install_and_set_up.yml
```diff
@@ -28,17 +28,41 @@
...
+        - ansible_distribution == "RedHat"
+        - ansible_distribution_major_version == "7"
+    - name: Enable required repository for RHEL7 ARM arch
+      rhsm_repository:
+        name: rhel-7-server-rpms
+        state: enabled
+      when:
+        - ansible_architecture == "aarch64"
+        - ansible_distribution == "RedHat"
+        - ansible_distribution_major_version == "7"
```
kgaikwad, line 65: Here, using the same repository i.e rhel-7-server-rpms.
In case no additional changes required based on architecture, will it be good to have single condition based on distribution & major version?

kgaikwad (changes requested):

ansible-playbooks/ros_install_and_set_up.yml
apuntamb: done.

apuntamb:
Steps followed to test:

- Run the playbook.
- It fails if `insights-client` is not installed and `core_collect` is not enabled.
- Install `insights-client`. I had to remove a conflicting package called `redhat-access-insights` that did not allow installing `insights-client`. While removing the same, one of the dependencies called `abrt` might get removed too. Need to install `abrt` too.
- Set `core_collect=True`
- Run the playbook.
- `pmlogger` service should now be running.

ansible-playbooks/ros_install_and_set_up.yml
```diff
@@ -28,17 +28,32 @@
...
         name: rhel-8-for-x86_64-baseos-rpms,rhel-8-for-x86_64-appstream-rpms
         state: enabled
-      when: ansible_architecture == "x86_64"
-    - name: Enable required repository for ARM arch
+      when:
+        - ansible_architecture == "x86_64"
+        - ansible_distribution == "RedHat"
+        - ansible_distribution_major_version == "8"
+    - name: Enable required repository for RHEL8 ARM arch
       rhsm_repository:
```
kgaikwad, line 43: Here, it uses by default `rhsm_repository` command. 
I think, playbook should check whether system uses RHUI/RHSM repositories. 
Accordingly, it should use correct command to enable.

apuntamb: As per the slack conversation, we can modify this part in a separate ticket.

gnurag:
PR look good now, and works for me with RHEL-7 image from AWS Marketplace. I tried the following commands to register the VM with ROS.
```
subscription-manager config --rhsm.manage_repos=1
subscription-manager register
subscription-manager attach --pool xxxxxxxx
subscription-manager repos --enable rhel-7-server-ansible-2.8-rpms
yum install ansible
curl 'https://raw.githubusercontent.com/RedHatInsights/ros-backend/edd057ef5326ff7dccc41a6dffdab017a38054ba/ansible-playbooks/ros_install_and_set_up.yml' > ros_install.yml
vi /etc/ansible/hosts  # Add localhost at the end
vi /etc/insights-client/insights-client.conf # Enable core_collect = True
ansible-playbook -c local ros_install.yml 
insights-client --register
```

kgaikwad (approved):
Thank you @gnurag and @apuntamb for sharing testing result.
Overall looks good :+1: 
Merging these changes as @apuntamb is going to handle RHUI condition separately.

gnurag:
Tried the updated playbook on AWS RHEL-7.9.  `core_collect=True` is automatically appended at the end of `/etc/insights-client/insights-client.conf` file. After running `insights-client --register` it correctly shows up in the inventory.

LGTM

PR #86, opened 2021-06-30
RHIROS #188 Migrate to Clowder

apuntamb:


deployment.yaml
```diff
@@ -0,0 +1,109 @@
+---
```
kgaikwad, line 1: @apuntamb, what will be the expected value for `ENABLE_RBAC` in this case?

apuntamb: I will define the `ENABLE_RBAC` flag in the parameters with value `True`.

clowdapp.yaml
```diff
@@ -0,0 +1,116 @@
...
+            cpu: ${CPU_REQUEST}
+            memory: ${MEMORY_REQUEST}
+          limits:
+            cpu: ${CPU_LIMIT}
+            memory: ${MEMORY_LIMIT}
+        env:
+          - name: CLOWDER_ENABLED
+            value: ${CLOWDER_ENABLED}
+    database:
+      name: ros-db
```
BlakeHolifield, line 69: This can stay `ros` if you'd like, it will have "db" appended to it by Clowder.

clowdapp.yaml
```diff
@@ -0,0 +1,116 @@
...
+  spec:
+    envName: ${ENV_NAME}
+    dependencies:
+    - rbac
+    pods:
+      - name: backend
+        minReplicas: 1
+        image: ${IMAGE}:${IMAGE_TAG}
+        command: ["bash"]
+        args: ["-c", "python -m manage db upgrade && python -m ros.api.main"]
```
BlakeHolifield, line 20: I'm assuming this is a Django app? If so I've seen other apps break out their db migrations into init containers. If your app is setup to always handle `db upgrade`, then this is fine. Just a thought.

clowdapp.yaml
```diff
@@ -0,0 +1,116 @@
...
+            cpu: ${CPU_REQUEST}
+            memory: ${MEMORY_REQUEST}
+          limits:
+            cpu: ${CPU_LIMIT}
+            memory: ${MEMORY_LIMIT}
+        livenessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /api/ros/v0/status
+            port: 8080
```
BlakeHolifield, line 32: One thing to note is that your app needs to run on port `8000` now and your metrics will need to be exposed on `9000`. 

Docs for that decision are here: https://clouddot.pages.redhat.com/docs/dev/getting-started/migration/faq.html#_why_does_my_app_have_to_run_on_port_8000 

So you'll just need to update: https://github.com/RedHatInsights/ros-backend/blob/e056b7ac9b12bc334f185ad420b7c9bda4610f57/ros/api/main.py#L11 to be `8000`.

BlakeHolifield (changes requested):
Overall this looks great! You'll need to change the backend to serve on port `8000` instead of `8080`, but that's the only issue. Once that's done you should be good to go.

clowdapp.yaml
```diff
@@ -0,0 +1,123 @@
...
+          - name: CLOWDER_ENABLED
+            value: ${CLOWDER_ENABLED}
+          - name: ENABLE_RBAC
+            value: "${ENABLE_RBAC}"
+    - name: processor
+      podSpec:
+        image: ${IMAGE}:${IMAGE_TAG}
+        minReplicas: ${{MIN_REPLICAS}}
+        command: ["bash"]
+        args: ["-c", "python -m ros.processor.main"]
```
apuntamb, line 64: Hi @BlakeHolifield! Do we need `initContainers` here as well?

clowdapp.yaml
```diff
@@ -0,0 +1,123 @@
...
+  required: true
+  value: 1Gi
+- description: Maximum amount of CPU the build container can use.
+  displayName: CPU Limit
+  name: CPU_LIMIT
+  required: true
+  value: '1'
+- description: Is clowder enabled
+  name: CLOWDER_ENABLED
+  value: "False"
```
apuntamb, line 123: @BlakeHolifield Should this be set to `true` here?

apuntamb:
@BlakeHolifield as discussed over call, once all the queries are resolve, should we merge this PR?

clowdapp.yaml
BlakeHolifield: I'd leave it as "false" for now until you get everything migrated. We don't want to misconfigure the currently running instances.

ros/api/main.py
```diff
@@ -8,4 +8,4 @@
 initialize_routes(api)
 
 if __name__ == "__main__":
-    app.run(host='0.0.0.0', port=8080)
+    app.run(host='0.0.0.0', port=8000)
```
BlakeHolifield, line 11: Do you have any apps that depend on your service? Just making sure before you change the port here.

clowdapp.yaml
apuntamb: okay!

clowdapp.yaml
apuntamb: gotcha, thanks!

ros/api/main.py
apuntamb: Yeah so we have all these services running for us: https://github.com/RedHatInsights/ros-backend/blob/main/scripts/docker-compose.yml

apuntamb:
@kgaikwad Could you do a high level review of the PR, especially with respect to the port changes too?

scripts/docker-compose.yml
```diff
@@ -119,12 +119,12 @@ services:
     depends_on:
       - insights-inventory-mq
     ports:
-      - 8081:8080
+      - 8001:8000
```
kgaikwad, line 122: For ROS and here, you mentioned port value  `8000`, will it create any problem?

kgaikwad (changes requested):
@apuntamb, added one inline comment about port value.
I didn't test the changes.

scripts/docker-compose.yml
apuntamb: @kgaikwad I guess it slipped my mind that we have already modified the ports for concerned services in the file.
Could you try running the containers and ros api/processors once with this change, to see if it works for you well?

kgaikwad: Sure, I will verify it once. Is dev setup working fine on your local with mentioned ports change?

Makefile
```diff
@@ -8,18 +8,11 @@ metadata={"branch_info": {"remote_branch": -1, "remote_leaf": -1}, "bios_uuid":
 identity={"identity": {"account_number": "0000001","type": "User","user": {"username": "tuser@redhat.com","email": "tuser@redhat.com","first_name": "test","last_name": "user","is_active": true,"is_org_admin": false, "is_internal": true, "locale": "en_US"},"internal": { "org_id": "000001"}}}
 b64_identity=$(shell echo '${identity}' | base64 -w 0 -)
 
-
-ros-upload-data:
-	curl -vvvv -F "upload=@sample-files/ros-sample.tar.gz;type=application/vnd.redhat.resource-optimization.sample+tgz" -F 'metadata=${metadata}' \
-		-H "x-rh-identity: ${b64_identity}" \
-		-H "x-rh-request_id: testtesttest" \
-		localhost:3000/api/ingress/v1/upload  | python -m json.tool
-
```
apuntamb, line 17: I hope we no longer need this command, right @kgaikwad ?

scripts/docker-compose.yml
apuntamb: Yeah its working in my local with the ports change :)

kgaikwad (approved):
Tested docker-compose file change. 
I can see all required services up and running. 
Tested archive upload and found data stored correctly into DB.
No further comments from my end. 

@BlakeHolifield - do you have any additional comments?

clowdapp.yaml
```diff
@@ -0,0 +1,123 @@
...
+    deployments:
+    - name: backend
+      podSpec:
+        image: ${IMAGE}:${IMAGE_TAG}
+        minReplicas: 1
+        initContainers:
+          - command:
+            - sh
+            - -c
+              python -m manage db upgrade && python -m ros.api.main
```
BlakeHolifield, line 24: I think you need an extra `-` here.

apuntamb: But, isn't it something like this - https://github.com/RedHatInsights/vulnerability-engine/blob/master/deploy/clowdapp.yaml#L230-L235

BlakeHolifield: Right. You can do it that way as well, but I don't think the current version works without that `- >-` line before the python code.

clowdapp.yaml
```diff
@@ -0,0 +1,123 @@
...
+    deployments:
+    - name: backend
+      podSpec:
+        image: ${IMAGE}:${IMAGE_TAG}
+        minReplicas: 1
+        initContainers:
+          - command:
+            - sh
+            - -c
+            - python -m manage db upgrade && python -m ros.api.main
```
apuntamb, line 24: @BlakeHolifield Is this the correct syntax?

kgaikwad:
From the offline discussion with @apuntamb, this pull-request is ready to go. So merging it.
Thank you @BlakeHolifield for review and @apuntamb  for addressing all comments.

PR #103, opened 2021-09-02
RHIROS-274 - Adding stale_timestamp column in the systems table

r14chandra:


ros/lib/models.py
```diff
@@ -55,6 +55,7 @@ class System(db.Model):
     rule_hit_details = db.Column(JSONB)
     state = db.Column(db.String(25))
     number_of_recommendations = db.Column(db.Integer)
+    stale_timestamp = db.Column(db.String(50))
```
apuntamb, line 58: @r14chandra  Since it is a Date type field, we should use a different type to store the value.
```suggestion
    stale_timestamp = db.Column(db.DateTime)
```
@kgaikwad do we need to provide any default value in case of `stale_timestamp`?

apuntamb (changes requested):
Overall looks good. Added a minor comment inline.

ros/lib/models.py
kgaikwad: @apuntamb - No default value is required here.

migrations/versions/e67b2b19df0e_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'e67b2b19df0e'
+down_revision = '0cc86f3e4037'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('systems', sa.Column('stale_timestamp', sa.String(length=50), nullable=True))
```
kgaikwad, line 21:
```suggestion
    op.add_column('systems', sa.Column('stale_timestamp', sa.DateTime(timezone=True), nullable=True))
```

ros/lib/models.py
```diff
@@ -55,6 +55,7 @@ class System(db.Model):
     rule_hit_details = db.Column(JSONB)
     state = db.Column(db.String(25))
     number_of_recommendations = db.Column(db.Integer)
+    stale_timestamp = db.Column(db.DateTime)
```
apuntamb, line 58: If you're setting `timezone=True` in the migration, then it should come from the model.
```suggestion
    stale_timestamp = db.Column(db.DateTime(timezone=True))
```

apuntamb (approved):
LGTM! :+1:

apuntamb:
Merged, thanks @r14chandra ! :100:

PR #104, opened 2021-09-06
RHIROS-270 - new is_configured API to check ROS systems exist

kgaikwad:


ros/api/v0/hosts.py
```diff
@@ -72,6 +72,11 @@ def get(self):
         # Refer - https://www.postgresql.org/docs/13/queries-limit.html
 
         account_query = db.session.query(RhAccount.id).filter(RhAccount.account == ident['account_number']).subquery()
+        system_count = db.session.query(System.id)\
+            .filter(System.account_id.in_(account_query)).count()
+        if system_count <= 0:
+            abort(404)
```
kgaikwad, line 78: I guess this response status code might conflict with other implementations( for example - RBAC). 
I need to figure out if  any other insights service is having similar scenario and uses any particular status code? OR suitable response code to avoid conflict.

kgaikwad: I have removed `404` handling inside `/systems` API and introduced new API `is_configured`.

ros/api/v0/hosts.py
```diff
@@ -13,12 +13,35 @@
...
+                'success': False,
+                'code': 'NOSYSTEMS',
+                'count': system_count
+            }
+
+        return {
+            'success': True,
+            'code': 'SYSTEMSEXIST',
+            'count': system_count
+        }
```
apuntamb, line 42: Can we instead use an if-else block with a single return statement? Either way works.

```suggestion
        if system_count <= 0:
            success = False
            systems_code = 'NOSYSTEMS'
        else:
            success = True
            systems_code = 'SYSTEMSEXIST'

        return {
            'success': success,
            'code': systems_code,
            'count': system_count
        }       
```

kgaikwad: @apuntamb,
Thank you for review. 
Here, I would prefer to return directly `dict` objects instead of declaring an extra variables.  
In case where it requires multiple keys & operations to perform based on condition & on this dict object before returning final result then definitely I would have preferred `if-else` loop.
Any thoughts? otherwise I will go ahead and modify the change as per your suggestion.

apuntamb: Yeah sure, makes sense! We can go ahead with the current implementation.

apuntamb (approved):
LGTM! :+1:

apuntamb:
Tested locally, works as expected. Merging, thanks @kgaikwad !

PR #106, opened 2021-09-15
RHIROS-293 - rename score to utilization

kgaikwad:


ros/api/v0/hosts.py
```diff
@@ -26,9 +26,14 @@
 
 class HostsApi(Resource):
     display_performance_score_fields = {
-        'cpu_score': fields.Integer,
-        'memory_score': fields.Integer,
-        'io_score': fields.Integer
+        'cpu': fields.Integer,
+        'memory': fields.Integer,
+        'io': fields.Integer
+    }
```
apuntamb, line 32: @kgaikwad Do we still need `display_performance_score_fields`?

ros/api/v0/hosts.py
```diff
@@ -37,6 +42,7 @@ class HostsApi(Resource):
         'account': fields.String,
         'number_of_recommendations': fields.Integer,
         'state': fields.String,
+        'performance_utilization': fields.Nested(performance_utilization_fields),
         'display_performance_score': fields.Nested(display_performance_score_fields),
```
apuntamb, line 46: ditto

ros/lib/models.py
```diff
@@ -23,7 +23,7 @@ class PerformanceProfile(db.Model):
     @property
     def display_performance_score(self):
         display_performance_score = {}
-        for key, value in self.performance_score.items():
+        for key, value in self.performance_utilization.items():
             score = value // 20
             display_performance_score[key] = 1 if score < 1 else score
```
apuntamb, line 29: Let's rename `display_performance_score` to something like `display_performance_utilization` term?

kgaikwad: One thing about `display_performance_score` field is that it contains 1 to 5 numbers for keys (cpu/io/memory). Do you still want to rename this to utilization?

ros/api/v0/hosts.py
kgaikwad: Yes, as mentioned offline while implementation frontend will need these score values to decide colour code.
If we want to remove this field then frontend needs to compute score to apply colour codes per row.

@apuntamb, I think, this field will get removed later based on upcoming colour code changes.

apuntamb: Gotcha, sure thing!

ros/lib/models.py
apuntamb: gotcha, since we are keeping the score.

kgaikwad:
@apuntamb & @r14chandra, do you have any additional comments? 
Did you try running these migration locally?

r14chandra:
@kgaikwad Tested locally, it's working now.

apuntamb:
Merged! Thanks @kgaikwad and @r14chandra!

PR #107, opened 2021-09-15
RHIROS-299 Filter systems by state

r14chandra:


ros/api/v0/hosts.py
```diff
@@ -101,11 +102,20 @@ def get(self):
...
                 .filter(System.display_name.ilike(f'%{filter_display_name}%'))\
                 .filter(System.account_id.in_(account_query))\
                 .filter(System.state.in_(SYSTEM_STATES_EXCEPT_EMPTY))
+        elif filter_by_state:
+            system_query = db.session.query(System.id)\
+                .filter(System.state.ilike(f'%{filter_by_state}%'))\
+                .filter(System.account_id.in_(account_query))
         else:
             system_query = db.session.query(System.id)\
                 .filter(System.account_id.in_(account_query))\
```
kgaikwad, line 121: Thank you @r14chandra for adding state filter.
Functionality wise looks good but I think we need to refactor this `if..elif..else` logic.
Currently, we just have three filters - account, display_name & state, think about adding more filters. 
@r14chandra -  To make it easier to add new filter as well as more readable, it would be great if you could refactor this code block.

kgaikwad (changes requested):

ros/api/v0/hosts.py
```diff
@@ -153,6 +152,11 @@ def get(self):
             limit, offset, hosts, count
         )
 
+    # Filter systems
+    def filter_with(self, filter_name, filter_param, system_query):
+        return db.session.query(System.id).filter(System.id.in_(system_query.subquery()))\
+                .filter(System.__dict__[filter_name].ilike(f'%{filter_param}%'))
```
kgaikwad, line 158: +1, it's good to have a separate method for filtering. You are adding one more nested subquery but I would prefer to apply filters with same system query.
How about having a method that just build system related filters & apply those using filter method in main system_query?

Example - 

```
system_query = db.session.query(System.id).filter(
            System.account_id.in_(account_query)
        ).filter(*self.build_system_filters())
```

`build_system_filters` is static method which responsible to build system related filters as follows -
  ```
  @staticmethod
    def build_system_filters():
        """Build system filters."""
        filters = []
        if (filter_display_name := request.args.get('display_name')):
            filters.append(
                System.display_name.ilike(f'%{filter_display_name}%'))

        if (filter_by_state := request.args.get('state')):
            filters.append(System.state.ilike(f'%{filter_by_state}%'))
        else:
            filters.append(System.state.in_(SYSTEM_STATES_EXCEPT_EMPTY))

        return filters


```

@r14chandra - what do you think?  Will it be easy for one to add extra system filter using this method?

(Note - you can rename `build_system_filters` to something else I just provided randomly. :))

r14chandra: @kgaikwad Yeah, we can do that. Adding this `build_system_filters` method in the code.

kgaikwad (approved):
@r14chandra, thank you for refactoring code. 
LGTM :+1: :ok_hand: 

Added new enhancement ticket [RHIROS-321](https://issues.redhat.com/browse/RHIROS-321)
for future.

PR #108, opened 2021-09-25
Remove DB section from cdappconfig

apuntamb:
Currently, there's no JIRA associated with this PR as it is based on the security alarm raised because of the password present in this file. Although, it is a dummy password, it would be wise to remove it from the file. Secondly, according to https://consoledot.pages.redhat.com/docs/dev/getting-started/migration/config.html#_cdappconfig, this file is generated by clowder. So, not sure if we do want to create it explicitly in our repo.

Based on further discussions, will create a corresponding JIRA as required.

### DIFF
--- a/cdappconfig.json
+++ b/cdappconfig.json
@@ -1,13 +1,4 @@
 {
-    "database": {
-        "adminPassword": "QM2FppyaTcYBcXtl",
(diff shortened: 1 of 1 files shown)


kgaikwad:
Thanks @apuntamb, merging it.

PR #109, opened 2021-09-28
RHIROS #331 - Fix puptoo & inv port in docker-compose.yml

apuntamb:


scripts/docker-compose.yml
```diff
@@ -119,12 +119,12 @@ services:
     depends_on:
       - insights-inventory-mq
     ports:
-      - 8001:8000
+      - 8001:8080
```
kgaikwad, line 122: @apuntamb - why we need `LISTEN_PORT` under `environment`?

kgaikwad (changes requested):

scripts/docker-compose.yml
apuntamb: I guess it was set earlier to override this default value here - https://github.com/RedHatInsights/insights-host-inventory/blob/0d0343e12da6e2343dc4ff03ea26739dd8f26a18/run_gunicorn.py#L10. But, looks like the port is already hardcoded here - https://github.com/RedHatInsights/insights-host-inventory/blob/master/Makefile#L18. And on this changed line i.e. 122, I am anyway forwarding the gunicorn's 8080 traffic on 8001. So, yeah makes sense to remove it. :+1:

kgaikwad (approved):
LGTM :+1:

PR #112, opened 2021-10-05
RHIROS-321 enhanced systems state filter

r14chandra:


apuntamb:
@r14chandra Could you rebase on top of the latest changes?

ros/api/v0/hosts.py
```diff
@@ -157,8 +157,8 @@ def build_system_filters():
         filters = []
         if filter_display_name := request.args.get('display_name'):
             filters.append(System.display_name.ilike(f'%{filter_display_name}%'))
-        if filter_by_state := request.args.get('state'):
-            filters.append(System.state.ilike(f'%{filter_by_state}%'))
+        if states := request.args.getlist('state'):
```
kgaikwad, line 160: Overall looks good to me. 
Just one concern here is that you are passing state filter list from query params directly to `in` clause. Will it be good to validate filter values against possible values of state?
@r14chandra, any thoughts?

r14chandra: @kgaikwad If we're using ROS Api Doc then I think we don't need as we can just select the mentioned **states** from the drop down but if we manually add filter to the url then we need to check.  What do you think?

PR #115, opened 2021-10-12
RHIROS 243 - Handle new 'Waiting for data' state from rules

apuntamb:
This is the first draft of the PR that assumes the following:

1. New rule change merged in insights-plugins would send the `Waiting for data` state key at the time of ros rule hits.
2. insights-core is the latest version on test machine.
3. Advisor engine rules are the latest version.

In addition to this, to track if the assumptions are as expected or negated, logged the state of the system while rule hits. In case the assumptions are negated, I will create a new Refs PR to handle the concerned scenario.

### DIFF
--- a/ros/processor/insights_engine_result_consumer.py
+++ b/ros/processor/insights_engine_result_consumer.py
@@ -81,16 +81,15 @@ class InsightsEngineResultConsumer:
                 account=host['account']
             )
 
-            state_key = reports[0].get('key')
-            LOG.info(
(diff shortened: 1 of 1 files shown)


kgaikwad:
@apuntamb - Is this pull-request ready for review?
If yes, could you please remove `WIP` tag from PR title?

kgaikwad (approved):
LGTM :+1: :ok_hand: 
Thank you @apuntamb for making log changes.

PR #122, opened 2021-10-28
RHIROS-366 | MOD - Removed display_performance_score_fields from API responses

saltgen:
Changes:
1. Removed definitions and usages of display_performance_score_fields

   Affected Endpoints:
       - `/api/ros/v0/systems`
       - `/api/ros/v0/systems/<host_id>`

2. Updated API definition(Swagger/OpenAPI) doc and README.md
3. Updated .gitignore
4. Initialized a couple of variables with None

Reasons:
1. Adoption of percentage scores instead of the score range i.e. 1-5
2. Updated the current response structure to maintain evenness
3. To exclude local IDE config fluff from getting git-ed
4. An attempt to keep flake8 happy

Concerns:

`HostHistoryApi` seems to have `display_performance_score_fields` data in 
it's 'data' key.

In case I remove the `display_performance_score_fields` I guess this endpoint
will get redundant.

Additional:

- All in-place tests have passed

- Will remove `models.PerformanceProfile.display_performance_score()` as well
  if display_performance_score_fields need to get removed from HostHistoryApi

- Did not add any new tests as the mentioned fields were not being consumed on front-end in any case.

- Responses have been tested with the current/new archives

- Sample response:

```
{
    "fqdn": "ip-172-31-25-176.ec2.internal",
    "inventory_id": "cb973a84-8607-4249-9395-6d0137d1ca72",
    "display_name": "ip-172-31-25-176.ec2.internal",
    "performance_utilization": {
        "cpu": 0,
        "memory": 0,
        "io": 0
    },
    "rating": 0,
    "number_of_suggestions": 0,
    "state": "Optimized",
    "report_date": "2021-10-28",
    "instance_type": "t2.micro",
    "cloud_provider": "aws",
    "idling_time": null,
    "io_wait": null
}
```


.gitignore
```diff
@@ -133,3 +133,7 @@ dmypy.json
 
 scripts/minio-conf/
 scripts/minio-data/
+
+# Editor/Local
```
kgaikwad, line 137: I guess, you can make use of personal git ignore rules?
For reference - https://www.atlassian.com/git/tutorials/saving-changes/gitignore#personal-git-ignore-rules

kgaikwad (changes requested):
Thank you @saltgen! 
Added one inline comment otherwise LGTM  :+1: :ok_hand:

apuntamb:
Hi @saltgen ! Thanks for the PR. Looks great !
I just had one question. Why can't we show `performance_utilization_fields` instead and remove the `display_performance_score_fields` from the `HostHistoryAPI`? Wouldn't that make more sense or is there any specific reason of not using those fields here that I'm missing? 
Thoughts @kgaikwad ?

.gitignore
saltgen: Personal git configs :ok_hand: Actually thanks for the tip :raised_hands: 

I feel we can still have these in our .gitignore just in case someone is new to git and makes their first contribution on this repo.

This way a lot of junk won't be able to make it through.

kgaikwad:
> Why can't we show `performance_utilization_fields` instead and remove the `display_performance_score_fields` from the `HostHistoryAPI`? 

+1 for removing `display_performance_score_fields` from the `HostHistoryAPI` as well and add `performance_utilization_fields`.

.gitignore
kgaikwad: Okay, @saltgen! 
Will it be possible for you to raise a separate PR for gitignore change? as these changes are not relevant to original task. 
Regardless of how big or tiny change it is, I would prefer to open a separate pull-request. In my opinion it is easy to move such change quickly rather than blocking it due to original task.

saltgen: Yes I agree, sounds good!
I will add in a few more exclusions for the ignore file on the new PR.

saltgen: Done! PR [here](https://github.com/RedHatInsights/ros-backend/pull/127)

saltgen:
> > Why can't we show `performance_utilization_fields` instead and remove the `display_performance_score_fields` from the `HostHistoryAPI`?
> 
> +1 for removing `display_performance_score_fields` from the `HostHistoryAPI` as well and add `performance_utilization_fields`.

Changes pushed :+1: 

Additionally I see this is per system/host endpoint, /systems/`<host_id>`/history.

Since there is code present for pagination, can we expect multiple performance_profiles per system?

.gitignore
```diff
@@ -132,4 +132,4 @@ dmypy.json
 .pyre/
 
 scripts/minio-conf/
-scripts/minio-data/
+scripts/minio-data/
```
kgaikwad, line 135: nitpick - could you please remove this file completely from commit?

kgaikwad (changes requested):
Thank you @saltgen for removing function from model and fixing history API.
Legitimate except a small nitpick!

.gitignore
saltgen: Added a return character at the end :+1:

kgaikwad:
> Since there is code present for pagination, do we expect multiple performance_profiles per system?

Yes, there is an archive upload every time when the insights-client runs on client system.

kgaikwad (approved):

kgaikwad:
@apuntamb - any additional comments?

apuntamb:
> @apuntamb - any additional comments?

Nope, looks good now. I haven't tested the workflow but codewise looks like nothing might break. ACK from my side.

kgaikwad:
Thank you @saltgen :+1: :tada:

PR #126, opened 2021-11-10
RHIROS-357 | REMOVE - OpenAPI servers drop-down has template string

saltgen:
Changes:
1. Removed servers object from openapi.json file

Reasons:
1. Getting auto-magically generated; checked on ci.foo and qa.foo via local setup

Issue:
The template string for the servers object was appearing as the default value on the drop-down for Servers:
Hence, the test calls were failing.

### DIFF
--- a/ros/openapi/openapi.json
+++ b/ros/openapi/openapi.json
@@ -10,11 +10,6 @@
         },
         "version": "v0.1"
     },
-    "servers": [
-        {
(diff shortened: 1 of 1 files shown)


r14chandra (approved):
Tested locally. LGTM! 👍🏻 @saltgen

kgaikwad (approved):

kgaikwad:
Thank you @saltgen and @r14chandra!

PR #129, opened 2021-11-11
Refs RHIROS-385 - fixes host keyerror

kgaikwad:
With this commit, it will show correct values for host id and account.
Message object is different for create/updated and delete event.
It seems that these keys were introduced quite sometime ago but issue never caught before.

### DIFF
--- a/ros/processor/inventory_events_consumer.py
+++ b/ros/processor/inventory_events_consumer.py
@@ -50,9 +50,19 @@ class InventoryEventsConsumer:
             if msg.error():
                 print(msg.error())
                 raise KafkaException(msg.error())
+
+            account = None
+            host_id = None
             try:
                 msg = json.loads(msg.value().decode("utf-8"))
                 event_type = msg['type']
+                if event_type == 'delete':
+                    account = msg['account']
+                    host_id = msg['id']
+                else:
+                    account = msg['host']['account']
+                    host_id = msg['host']['id']
+
                 if event_type in self.event_type_map.keys():
                     handler = self.event_type_map[event_type]
                     handler(msg)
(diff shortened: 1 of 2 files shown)


kgaikwad:
Reference doc - https://consoledot.pages.redhat.com/docs/dev/services/inventory.html#_delete_event

ros/processor/inventory_events_consumer.py
```diff
@@ -97,21 +107,25 @@ def host_delete_event(self, msg):
...
-            db.session.commit()
 
     def host_create_update_events(self, msg):
         """ Process created/updated message ( create system record, store new report )"""
         self.prefix = "PROCESSING Create/Update EVENT"
         if 'is_ros' in msg['platform_metadata']:
+            LOG.info(
+                'Process a msg for host(%s) belonging to account %s - %s',
+                msg['host']['id'], msg['host']['account'], self.prefix
+            )
```
apuntamb, line 128: I'm not sure IIUC, but does this log msg want to convey something like this:
```suggestion
            LOG.info(
                'Processing inventory events message for host(%s) belonging to account %s - %s',
                msg['host']['id'], msg['host']['account'], self.prefix
            )
```

apuntamb (commented):
Thanks for the quick fix @kgaikwad !! :100:  Just added a minor query. Rest it's good to go.

apuntamb:
> Reference doc - https://consoledot.pages.redhat.com/docs/dev/services/inventory.html#_delete_event

This was really helpful! Thanks for sharing.

ros/processor/inventory_events_consumer.py
kgaikwad: As we are already adding a prefix into message, skipping `inventory events` substring. Except inventory, ROS is not listening to any other service for create/updated event. I hope this fine!

apuntamb (approved):
lgtm :+1:

apuntamb:
Merged, thanks for the fix @kgaikwad !

PR #130, opened 2021-11-15
RHIROS #387 - Increase kafka consumer timeouts

apuntamb:
Based on the kafka logs as described in the JIRA ticket, bumping the timeout values up would be a good test.
Reference doc - https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#consumerconfigs_heartbeat.interval.ms


apuntamb:
Closing this as there was no similar issue observed and no action item set. Will reopen this PR when required.

PR #135, opened 2021-11-24
Refs #RHIROS 367 - fixed file name in the makefile for uploading data

PreetiW:
### Background

currently if we run the `make insights-upload-data` command it is failing due to file name error as we recently updated the sample files for uploading data. This PR fixes the file name for same.


apuntamb (approved):
Good catch @PreetiW ! Looks good. :100:

apuntamb:
Merged @PreetiW ! Thanks for the quick fix.

PR #136, opened 2021-11-24
revert metrics port change

kgaikwad:


apuntamb (approved):
Thanks for fixing this @kgaikwad ! As discussed, this port is not for clowder config. So, we can keep it as is to avoid port conflicts in local setup as well.

apuntamb:
Merged, thanks @kgaikwad !

PR #138, opened 2021-12-02
RHIROS-402 | MOD - Move utilization logic from inventory processor to engine processor

saltgen:
### Changes:
1. Engine consumer grabs performance_utilization from msg, creates performance profiles as well
2. Events consumer no longer creates performance profiles, calculation/aggregation scrapped
3. Updated process_archive.py with custom log parameters
4.  Added an util function for validating booleans, `validate_type`

### Reasons:
1. Avoiding duplication since rules are now available from engine topic :rabbit2: 
2. Moved performance profile code to engine consumer, functions served no more purpose
3. Developer friendly logging :v:
4. Since our processing depends on a foreign service(advisor, puptoo) the backend needs to handle booleans smartly

### Additional:
* Sample producer with _supreme omnipotent intelligence_ for testing [here](https://pastebin.com/1HLxivWH) :wink: 
*  Will need to do a `pipenv install kafka-python` though

**Note**: The engine msg that gets generated from the latest archive, _new-metrics-test-archive.tar.xz_ shared with team 
at 3:31 PM 1st Dec 2021 will not work since it has `"_score"` keys.
        
```
msg["results"]["system"]["metadata"]{
     ...
    "cpu_score": 1,
    "mem_score": 4,
    "io_score": 1,
    ...
}
```


ros/processor/insights_engine_result_consumer.py
```diff
@@ -116,6 +130,26 @@ def process_report(self, host, reports):
                     number_of_recommendations=rec_count,
                     state=SYSTEM_STATES[state_key]
                 )
+                LOG.info(
+                    f"{self.prefix} - System created/updated successfully: {host['id']}"
```
apuntamb, line 134: For consistency, let's either use `self.prefix` or `custom_prefix` globally.

ros/processor/utils.py
```diff
@@ -0,0 +1,17 @@
...
+    :param value: Value to validate.
+    :param type_: Type to validate against.
+    :return: True if the value is of the specified type, False otherwise.
+    """
+    if type_ == bool:
+        # ast.literal_eval does not understand lowercase 'True' or 'False'
+        value = value.capitalize() if value in ['true', 'false'] else value
+    evaluated_value = type_evaluation.literal_eval(value) if value else None
+
+    return True if type(evaluated_value) == type_ else False
```
apuntamb, line 17: Can we define this method in https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py ? This would avoid some overhead as it already contains couple of helper methods.

ros/processor/insights_engine_result_consumer.py
```diff
@@ -116,6 +130,26 @@ def process_report(self, host, reports):
...
+                )
+
+                performance_utilization = {
+                    'memory': system_details['mem_utilization'],
+                    'cpu': system_details['cpu_utilization'],
+                    'io': system_details['io_utilization']
+                }
```
apuntamb, line 141: Can you point me to the logic where you're handling the Waiting for Data state i.e. the 0 day experience where you won't have any actual metric values available?

apuntamb (changes requested):
Thanks for the thorough revision of the process @saltgen ! Added couple of comments inline, please take a look and let me know your thoughts on them.

ros/processor/insights_engine_result_consumer.py
saltgen: For such a case, won't the engine msg have 0 values? I'm not too sure

ros/processor/insights_engine_result_consumer.py
```diff
@@ -75,17 +78,28 @@ def run(self):
...
+                else []
+            ros_reports = [
+                report for report in reports
+                if 'cloud_instance_ros_evaluation' in report["rule_id"]
+            ]
+            self.process_report(host, ros_reports, system_metadata, performance_record)
 
-    def process_report(self, host, reports):
+    def process_report(self, host, reports, system_details, performance_record):
         """create/update system based on reports data."""
```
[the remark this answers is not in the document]
saltgen: Sorry missed this one :sweat_smile: Suggestion committed :+1:

ros/processor/insights_engine_result_consumer.py
```diff
@@ -75,17 +78,28 @@ def run(self):
                 self.consumer.commit()
 
     def handle_msg(self, msg):
-        if msg["input"]["platform_metadata"]["is_ros"]:
+        # Checking for match in a value list:
+        # An attempt to reduce dependency on foreign data, i.e. msg
```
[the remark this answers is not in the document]
saltgen: Yeah I do tend to get too verbose at times :facepalm: Suggestion committed :+1:

ros/processor/insights_engine_result_consumer.py
apuntamb: According to the older workflow, we were getting `NO_PCP_DATA` key hit from the rules and we had to handle it in our processor to store data in the db with no value scenario. I am guessing that should be the case even now, but I can confirm once I see the actual message that will come from the insights rule.

saltgen: I noticed that portion of the code and had removed it intentionally.
IMO we should receive the 0 values from the advisor for utilization in case its a fresh system.
Yes, lets wait for the rules to go live and then take a call, thanks!

ros/processor/utils.py
saltgen: I should have done that in the first place :facepalm: 
Thanks @apuntamb 
Changes pushed :rocket:

ros/processor/insights_engine_result_consumer.py
saltgen: The custom_prefix's default assignment was implemented such when `get_performance_profile` gets called
from the engine consumer, the logs messages start with `PROCESSING ENGINE RESULTS`
instead the archive processor's prefix i.e. `ARCHIVE PROCESSOR`

Changes [here](https://github.com/RedHatInsights/ros-backend/pull/138/files#diff-6255fd2253fb40ee086cf8feeb2c42da4450a36667cd5e6405801fe4e81e2e0fR59-R60)

In the commented file's case the self.prefix seems to do its job, or am I missing something here :sweat_smile:

apuntamb: So IIUC and based on a quick look at the code, we do want the log to have `ARCHIVE PROCESSOR` prefix here as that is where these log statements reside.
```python
2021-10-14 19:40:51,195 - INFO  - get_performance_profile - ARCHIVE PROCESSOR - Downloading the report from http://minio:9000/insights-upload-perma/<url continues with credentials>.
```

ros/processor/insights_engine_result_consumer.py
apuntamb: @saltgen As per Xiangce's latest comment on CEECBA-5875, no metadata would be returned. Hence, we will have to set `*_utilization` values to 0 as present in the red diff [here](https://github.com/RedHatInsights/ros-backend/pull/138/files#diff-3ec10f47264a06a134d0bf1ce745fb26ad086de0af6da0cf6e83c00dcf2c4193L156). We should check if the for the NO_PCP_DATA condition and accordingly set the values.

saltgen: I see, okay, I will make the changes and push shortly.

saltgen: As discussed, awaiting live rules and archives to test changes.
Will be adding suggested change in one go.

ros/processor/insights_engine_result_consumer.py
saltgen: As discussed on Slack, keeping this change in place.

ros/processor/insights_engine_result_consumer.py
```diff
@@ -114,7 +127,28 @@ def process_report(self, host, reports):
                     fqdn=host['fqdn'],
                     rule_hit_details=reports,
                     number_of_recommendations=rec_count,
-                    state=SYSTEM_STATES[state_key]
+                    state=SYSTEM_STATES[state_key],
+                    instance_type=performance_record.get('instance_type')
```
saltgen, line 131: @apuntamb FYI, added as per conversation, https://github.com/RedHatInsights/ros-backend/pull/141#discussion_r765536954

kgaikwad (changes requested):
@saltgen, added two small comments.

ros/processor/insights_engine_result_consumer.py
```diff
@@ -75,18 +78,28 @@ def run(self):
                 self.consumer.commit()
 
     def handle_msg(self, msg):
-        if msg["input"]["platform_metadata"]["is_ros"]:
+        # Checking for boolean value in is_ros key
```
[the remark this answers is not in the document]
saltgen: Done, change pushed :+1:

ros/processor/insights_engine_result_consumer.py
```diff
@@ -75,18 +78,28 @@ def run(self):
...
+                msg['input']['platform_metadata']['account'],
+                custom_prefix=self.prefix
+            )
+            reports = msg["results"]["reports"]  \
+                if msg["results"]["reports"] \
+                and type(msg["results"]["reports"]) == list \
+                else []
+            ros_reports = [
+                report for report in reports
+                if 'cloud_instance_ros_evaluation' in report["rule_id"]
```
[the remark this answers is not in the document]
saltgen: Change committed

ros/processor/insights_engine_result_consumer.py
```diff
@@ -114,7 +126,28 @@ def process_report(self, host, reports):
...
+                    instance_type=performance_record.get('instance_type')
+                )
+                LOG.info(
+                    f"{self.prefix} - System created/updated successfully: {host['id']}"
+                )
+
+                performance_utilization = {
+                    'memory': system_details['mem_utilization'],
+                    'cpu': system_details['cpu_utilization'],
+                    'io': system_details['io_utilization']
```
apuntamb, line 139: @saltgen Good choice on keeping `system_details` consistent with `system_metadata`. But, would it make more sense either to keep `system_metadata` as is or rename it to `performance_details` or `performance_*`. The reason being if we keep it `system_details`, it should not be misinterpreted as any field or details of our `systems` table in ROS. Let me know what you think. Example added as a suggestion below.
```suggestion
                performance_utilization = {
                    'memory': system_metadata['mem_utilization'],
                    'cpu': system_metadata['cpu_utilization'],
                    'io': system_metadata['io_utilization']
```

ros/processor/insights_engine_result_consumer.py
apuntamb: Also, you will need to handle the Optimized state here too, otherwise it could give an error as follows:
```python
2021-12-14 13:18:10,504 - INFO  - process_report - PROCESSING ENGINE RESULTS - No ROS rule hits found for system with inventory id: cd002de7-851e-40af-a70e-24554ab10b55. Hence, marking the state as Optimized.
2021-12-14 13:18:10,506 - INFO  - process_report - PROCESSING ENGINE RESULTS - System created/updated successfully: cd002de7-851e-40af-a70e-24554ab10b55
2021-12-14 13:18:10,506 - ERROR  - process_report - PROCESSING ENGINE RESULTS - Unable to add host ip-172-31-45-168.ap-south-1.compute.internal to DB belonging to account: 0000001 - 'mem_utilization'

```

ros/processor/insights_engine_result_consumer.py
```diff
@@ -114,7 +126,28 @@ def process_report(self, host, reports):
...
+
+                get_or_create(
+                    db.session, PerformanceProfile, ['system_id', 'report_date'],
+                    system_id=system.id,
+                    performance_record=performance_record,
+                    performance_utilization=performance_utilization,
+                    report_date=datetime.datetime.utcnow().date()
+                )
+                LOG.info(
+                    f"{self.prefix} - Performance profile created successfully: {host['id']}"
```
apuntamb, line 150:
```suggestion
                    f"{self.prefix} - Performance profile created successfully for the system: {host['id']}"
```

apuntamb: @saltgen Apart from your commits, can you update [this line](https://github.com/RedHatInsights/ros-backend/pull/138/files#diff-0c17bcf0cdabfe68b39e608007e0195be1c3ae63f6395c4f9bd809e003089e26R164) replacing `host['fqdn']` to `host['id']` or another relevant key to display the inventory id and not fqdn since in all the logs elsewhere we are referring to the inventory id and not the fqdn. So, to keep the logging helpful to debug and consistent, I would suggest replacing the key.

apuntamb (changes requested):
Almost there, just added few minor suggestions worth looking at. :)

ros/processor/insights_engine_result_consumer.py
saltgen: Sure, change pushed

saltgen: Sure, change pushed

ros/processor/insights_engine_result_consumer.py
saltgen: Thank you!
Yeah makes sense, updated `system_metadata` to `utilization_info`
Changes pushed :+1:

ros/processor/insights_engine_result_consumer.py
saltgen: > @saltgen As per Xiangce's latest comment on CEECBA-5875, no metadata would be returned. Hence, we will have to set `*_utilization` values to 0 as present in the red diff [here](https://github.com/RedHatInsights/ros-backend/pull/138/files#diff-3ec10f47264a06a134d0bf1ce745fb26ad086de0af6da0cf6e83c00dcf2c4193L156). We should check for the NO_PCP_DATA condition and accordingly set the values.

Added zero_value initialization for NO_PCP_DATA. Changes pushed :rocket: 



> Also, you will need to handle the Optimized state here too, otherwise it could give an error as follows:
> 
> ```python
> 2021-12-14 13:18:10,504 - INFO  - process_report - PROCESSING ENGINE RESULTS - No ROS rule hits found for system with inventory id: cd002de7-851e-40af-a70e-24554ab10b55. Hence, marking the state as Optimized.
> 2021-12-14 13:18:10,506 - INFO  - process_report - PROCESSING ENGINE RESULTS - System created/updated successfully: cd002de7-851e-40af-a70e-24554ab10b55
> 2021-12-14 13:18:10,506 - ERROR  - process_report - PROCESSING ENGINE RESULTS - Unable to add host ip-172-31-45-168.ap-south-1.compute.internal to DB belonging to account: 0000001 - 'mem_utilization'
> ```

As discussed on Slack since metadata is unavailable only in-case of `NO_PCP_DATA`, letting this be :v:

ros/processor/insights_engine_result_consumer.py
```diff
@@ -1,19 +1,22 @@
...
                                    kafka_failures)
+from ros.processor.process_archive import get_performance_profile
+from ros.lib.utils import validate_type
 
 SYSTEM_STATES = {
     "INSTANCE_OVERSIZED": "Oversized",
     "INSTANCE_UNDERSIZED": "Undersized",
     "INSTANCE_IDLE": "Idling",
-    "INSTANCE_UNDER_PRESSURE": "Under pressure",
+    "INSTANCE_OPTIMIZED_UNDER_PRESSURE": "Under pressure",
```
saltgen, line 19: @apuntamb Noticed the rule hit had this key, INSTANCE_OPTIMIZED_UNDER_PRESSURE, hence the change

ros/processor/insights_engine_result_consumer.py
```diff
@@ -115,7 +128,42 @@ def process_report(self, host, reports):
...
+                )
+                LOG.info(
+                    f"{self.prefix} - System created/updated successfully: {host['id']}"
+                )
+
+                # For Optimized state, reports would be empty, but utilization_info would be present
+                if reports:
+                    set_default_utilization = True if reports[0].get('key') == 'NO_PCP_DATA' else False
+                else:
+                    set_default_utilization = False
```
saltgen, line 142: @apuntamb This was the only new change post our discussions, hope I'm making sense here

ros/processor/insights_engine_result_consumer.py
```diff
@@ -115,7 +128,42 @@ def process_report(self, host, reports):
...
+
+                if set_default_utilization is False:
```
apuntamb, line 144: Can we check True condition first, if there is no specific reason for checking False first?
```suggestion
                if set_default_utilization:
```

ros/processor/insights_engine_result_consumer.py
apuntamb: Yeah +1

ros/processor/insights_engine_result_consumer.py
```diff
@@ -46,6 +49,7 @@ def __next__(self):
         return msg
 
     def run(self):
+        LOG.info("Processor is up, awaiting msgs ...")
```
apuntamb, line 52:
```suggestion
        LOG.info("%s Processor is running. Awaiting messages.", % self.prefix)
```

ros/processor/insights_engine_result_consumer.py
saltgen: The True will only apply for 'NO_PCP_DATA', hence if we evaluate False first then the processor code will get executed faster IMO.

apuntamb: +1
minor fix:
```suggestion
                if not set_default_utilization:
```

ros/processor/insights_engine_result_consumer.py
saltgen: Just tried this out, I see,

PROCESSING ENGINE **RESULTS** - Processor is running. Awaiting msgs.

Do we want the **RESULTS** part on this notifier log?

saltgen: Any better?

`2021-12-15 22:54:02,489 - INFO  - run - Processor is running. Awaiting msgs.`

apuntamb: So, currently, we do have all our logs with such prefixes. IMO, to keep debugging easier, we can use these prefixes so that we are aware the logs are generated from which processor.

ros/processor/insights_engine_result_consumer.py
saltgen: I might be missing the point here, but I feel `is` is more explicit, clearer and good for checking booleans.
`not` feels a little confusing for me cause of the following,

`>>> not False`
True
`>>> not 0` :thinking: 
True
`>>> not 1`
False
`>>> not {}` :thinking: 
True


Please do correct me if I'm wrong

apuntamb: Let's keep it as is for now.

apuntamb:
I would request @r14chandra to test this change end to end once in her local and we can then merge this right away.

ros/processor/insights_engine_result_consumer.py
saltgen: Set to _Processor is running. Awaiting messages._

saltgen: > So, currently, we do have all our logs with such prefixes. IMO, to keep debugging easier, we can use these prefixes so that we are aware the logs are generated from which processor

Okay, change pushed

ros/processor/insights_engine_result_consumer.py
```diff
@@ -115,7 +128,47 @@ def process_report(self, host, reports):
...
+                if set_default_utilization is False:
+                    performance_utilization = {
+                        'memory': utilization_info['mem_utilization'],
+                        'cpu': utilization_info['cpu_utilization'],
+                        'io': convert_iops_from_percentage(utilization_info['io_utilization'])
+                    }
+                    # Adding max_io to utilization dict
+                    performance_utilization.update(
+                       {'max_io': max(performance_utilization['io'].values())}
+                    )
```
saltgen, line 153: @apuntamb Added `max_io` to `performance_utilization` as we will need it as a sorting param in db queries as per changes requested [here](https://issues.redhat.com/browse/RHIROS-393).

apuntamb (approved):
Big change!! Thanks for addressing all the comments patiently @saltgen!! Worth bumping the API version soon. LGTM!! :)
Only testing missing here is for the Oversized state as the IOPS are not currently fetched due to the disabled IOPS functionality for now.

r14chandra (approved):
Tested locally with different state archives with my changes. It's working! We're good to go @saltgen @apuntamb

apuntamb:
@kgaikwad Any additional comments from your side?

ros/processor/insights_engine_result_consumer.py
```diff
@@ -115,7 +128,47 @@ def process_report(self, host, reports):
...
-                    state=SYSTEM_STATES[state_key]
+                    state=SYSTEM_STATES[state_key],
+                    instance_type=performance_record.get('instance_type')
+                )
+                LOG.info(
+                    f"{self.prefix} - System created/updated successfully: {host['id']}"
+                )
+
+                # For Optimized state, reports would be empty, but utilization_info would be present
+                if reports:
```
kgaikwad, line 139:
```suggestion
                set_default_utilization = False
                if reports and reports[0].get('key') == 'NO_PCP_DATA' :
                    set_default_utilization = True 

                if set_default_utilization is False:
                    performance_utilization = {
                        'memory': utilization_info['mem_utilization'],
                        'cpu': utilization_info['cpu_utilization'],
                        'io': convert_iops_from_percentage(utilization_info['io_utilization'])
                    }
                    # Adding max_io to utilization dict
                    performance_utilization.update(
                       {'max_io': max(performance_utilization['io'].values())}
                    )
                else:
                    LOG.info(f"{self.prefix} - Setting default utilization for performance profile")
                    performance_utilization = {
                        'memory': 0,
                        'cpu': 0,
                        'max_io': 0,
                        'io': {}
                    }
```
I can see that `set_default_utilization` is True only when key = `NO_PCP_DATA`. Will be it OK to set default value of `set_default_utilization` to False for easy understanding and readability?

kgaikwad: It is not allowing me to select multiple lines while adding comment.

ros/processor/insights_engine_result_consumer.py
```diff
@@ -115,7 +128,47 @@ def process_report(self, host, reports):
...
+                        'memory': utilization_info['mem_utilization'],
+                        'cpu': utilization_info['cpu_utilization'],
+                        'io': convert_iops_from_percentage(utilization_info['io_utilization'])
+                    }
+                    # Adding max_io to utilization dict
+                    performance_utilization.update(
+                       {'max_io': max(performance_utilization['io'].values())}
+                    )
+                else:
+                    LOG.info(f"{self.prefix} - Setting default utilization for performance profile")
```
kgaikwad, line 155:
```suggestion
                    LOG.debug(f"{self.prefix} - Setting default utilization for performance profile")
```
Do we really want to log this information with level `info`?

ros/processor/insights_engine_result_consumer.py
saltgen: @kgaikwad Thanks, changes tested and pushed :sunflower:

ros/processor/insights_engine_result_consumer.py
saltgen: Added comment on code as well

`# max_io will be used to sort systems endpoint response instead of io`

kgaikwad (approved):
LGTM :+1: 
Thank you @saltgen for this change.

apuntamb:
Merged, thanks @saltgen for the change and @r14chandra and @kgaikwad for reviewing thoroughly.

PR #137, opened 2021-11-30
RHIROS-401 Update archive processor to store new metrics

r14chandra:
Now the `profile` object in process_archive will contain new metrics with their respective values except `disk.dev.total` and `kernel.all` type metrics which will store its value as an object. This is because we're not processing or using these metrics anywhere in our ros-backend. We're just storing in ros db as performance_record in `performance_profile` table.

Refer the below profile object.
`{
	"hinv.ncpu": 8.0,
	"total_cpus": 1,
	"mem.physmem": 32617072.0,
	"instance_type": "t2.micro",
	"disk.dev.total": {
		"nvme0n1": {
			"val": 7.22,
			"units": "count / sec"
		}
	},
	"mem.util.available": 27455175.254,
	"kernel.all.cpu.idle": 7.55,
	"kernel.all.pressure.io.full.avg": {
		"1 minute": {
			"val": 0.006,
			"units": "none"
		},
		"5 minute": {
			"val": 0.004,
			"units": "none"
		},
		"10 second": {
			"val": 0.012,
			"units": "none"
		}
	},
	"kernel.all.pressure.io.some.avg": {
		"1 minute": {
			"val": 0.006,
			"units": "none"
		},
		"5 minute": {
			"val": 0.004,
			"units": "none"
		},
		"10 second": {
			"val": 0.013,
			"units": "none"
		}
	},
	"kernel.all.pressure.cpu.some.avg": {
		"1 minute": {
			"val": 0.575,
			"units": "none"
		},
		"5 minute": {
			"val": 0.572,
			"units": "none"
		},
		"10 second": {
			"val": 0.579,
			"units": "none"
		}
	},
	"kernel.all.pressure.memory.full.avg": {
		"1 minute": {
			"val": 0.0,
			"units": "none"
		},
		"5 minute": {
			"val": 0.0,
			"units": "none"
		},
		"10 second": {
			"val": 0.0,
			"units": "none"
		}
	},
	"kernel.all.pressure.memory.some.avg": {
		"1 minute": {
			"val": 0.0,
			"units": "none"
		},
		"5 minute": {
			"val": 0.0,
			"units": "none"
		},
		"10 second": {
			"val": 0.0,
			"units": "none"
		}
	}
}`

### Testing
To test this change with an archive with new metrics it uses utilization and score values (which are currently using old metrics for calculation). For now set these values as zero in `inventory event processor` file for testing .

PR #139, opened 2021-12-02
Refs #RHIROS-401 - add Pipfile.lock

kgaikwad:
as new version of insights-core pkg is available.

+ @r14chandra into loop as this is relevant to RHIROS-401.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,10 +18,10 @@
     "default": {
         "alembic": {
             "hashes": [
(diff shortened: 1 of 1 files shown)


apuntamb:
@kgaikwad Thanks for the fix. Can you or @r14chandra confirm if it works fine on local and there's no conflict anywhere in the workflow?

r14chandra (approved):
@kgaikwad We can merge this PR 👍🏻 . #137 with these changes works fine.

PR #150, opened 2021-12-16
RHIROS-393 | io Utilization as dict and max_io

saltgen:
### Changes:
1. HostsApi modified marshall fields and added io dict sorting, max_io
2. HostDetailsApi modified marshall fields and added io dict sorting, max_io
   to keep it consistent with HostsApi
3. HostHistoryApi io values are now dict and sorted on values
4. max_io added to build_sort_expression for enabling ASC and DESC sorting
5. Added util function for sorting io_dict
6. Modified Make file for easier testing, can be reverted if deemed not necessary

### All Systems:
 `/api/ros/v0/systems?order_by=max_io&order_how=DESC`

```
"data": [
        {
            "fqdn": "rhel8-test.hyrdvbbl3ugefmkdahnvv2bx1h.rx.internal.cloudapp.net",
            "display_name": "rhel8-test.hyrdvbbl3ugefmkdahnvv2bx1h.rx.internal.cloudapp.net",
            "inventory_id": "f73715ea-00b6-4a02-a4aa-528584b04dfd",
            "account": "0000001",
            "number_of_suggestions": 0,
            "state": "Optimized",
            "performance_utilization": {
                "cpu": 100,
                "memory": 95,
                "max_io": 700.0,
                "io_all": {
                    "xvdb": 700,
                    "xvdc": 92,
                    "xvda": 72
                }
            },
            "cloud_provider": "azure",
            "instance_type": "Standard_B1s",
            "idling_time": "98.90",
            "io_wait": null
        },
        {
            "fqdn": "ip-172-31-45-168.ap-south-1.compute.internal",
            "display_name": "ip-172-31-45-168.ap-south-1.compute.internal",
            "inventory_id": "f2b4c41d-27bd-4785-9dc4-ac09b6782981",
            "account": "0000001",
            "number_of_suggestions": 0,
            "state": "Optimized",
            "performance_utilization": {
                "cpu": 5,
                "memory": 22,
                "max_io": 400.0,
                "io_all": {
                    "sdb": 400,
                    "sda": 100,
                    "xvga": 95
                }
            },
            "cloud_provider": "aws",
            "instance_type": "t2.micro",
            "idling_time": "99.60",
            "io_wait": null
        },
        {
            "fqdn": "ip-172-31-25-176.ec2.internal",
            "display_name": "ip-172-31-25-176.ec2.internal",
            "inventory_id": "88f2588b-a1f1-4638-8b45-50ca2c3ec93e",
            "account": "0000001",
[...]

PR #152, opened 2021-12-17
RHIROS-402 | Correction on util data types

saltgen:
### Changes:
1. Explicit convert of cpu, memory to int

### Additional:
* max_io and io dict value types are pending
* Reference: https://github.com/RedHatInsights/ros-backend/pull/150#issuecomment-996115794

### DIFF
--- a/ros/lib/utils.py
+++ b/ros/lib/utils.py
@@ -78,13 +78,18 @@ def validate_type(value, type_):
     return True if type(evaluated_value) == type_ else False
 
 
-def convert_iops_from_percentage(iops_dict):
+def convert_to_iops_actuals(iops_dict):
     """
-    Convert IOPS to MBPS
+    Convert IOPS to MBPS fom percentage.
+    In case IOPS values are actual,
+    converts them to int from str
     :param iops_dict: IOPS dict to convert.
     :return: IOPS values converted to MBPS.
     """
     iops_in_mbps = {}
     for key, value in iops_dict.items():
-        iops_in_mbps[key] = float(value) * 100
+        if float(value) < 1.0:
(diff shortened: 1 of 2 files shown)


kgaikwad (changes requested):
For `max_io` and IOPS value against each device, let's conside `int` only.

Referring  to - https://en.wikipedia.org/wiki/IOPS, I don't see any IOPS value in float there. 
@saltgen - what do you think?

saltgen:
> For `max_io` and IOPS value against each device, let's conside `int` only.
> 
> Referring to - https://en.wikipedia.org/wiki/IOPS, I don't see any IOPS value in float there. @saltgen - what do you think?

I referred to the same wiki for my conclusion, but from rules I am getting `io` as float, hence when I multiply it by 100
I get float

For instance `"io": {"xvda": 0.314}` will be saved up as 31.4, that said this is **from archive**.

Shall we still go ahead with `int` for  `max_io` and IOPS value?

kgaikwad:
> Shall we still go ahead with `int` for `max_io` and IOPS value?
Yes, it should be `int` for both.

> That said am I using the wrong multiplier i.e. 100. thinking Let me check with Pau
+1 for checking with Pau

saltgen:
> > Shall we still go ahead with `int` for `max_io` and IOPS value?
> 
> Yes, it should be `int` for both.
> 
> > That said am I using the wrong multiplier i.e. 100. thinking Let me check with Pau
> 
> +1 for checking with Pau

@kgaikwad Thanks, pushed changes here for the multiplier, in-place tests run fine, manually tested as well :+1:

kgaikwad (approved):
LGTM :+1:

PR #154, opened 2021-12-22
Refs RHIROS 133 - fix type error for metrics port

kgaikwad:
As metrics port value set as string in template, currently it is failing while restarting http server. 
To fix this, converting metrics port value to integer before passing as argument.

### DIFF
--- a/ros/processor/main.py
+++ b/ros/processor/main.py
@@ -29,7 +29,7 @@ if __name__ == "__main__":
     events.start()
     engine_results.start()
     collector.start()
-    start_http_server(METRICS_PORT)
+    start_http_server(int(METRICS_PORT))
     # Wait for threads to finish
     events.join()
     engine_results.join()


saltgen (approved):
We should analyze as why METRICS_PORT and [other env vars](https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/ros/ros-processor.yml)  are in str format on, but as of now this change will do. :+1:

kgaikwad:
> We should analyze as why METRICS_PORT and [other env vars](https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/ros/ros-processor.yml) are in str format on, but as of now this change will do. +1

@saltgen - could you please verify if integer type values are allowed for parameter value inside template? If it is allowed there, I will be happy to modify this value to int there. But this only applicable to current deployment.
In my opinion, it is always better to have conversion instead of relying on upcoming values & their format.

saltgen:
> > We should analyze as why METRICS_PORT and [other env vars](https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/ros/ros-processor.yml) are in str format on, but as of now this change will do. +1
> 
> @saltgen - could you please verify if integer type values are allowed for parameters inside template? If it is allowed there, I will be happy to modify this value to int there. In my opinion, it is always better to have conversion instead of relying on upcoming values & their format.

Yes agreed, I'll check and revert on the same.

PR #155, opened 2021-12-23
RHIROS #448 - Make IsConfigured API response consistent with Openapi spec

apuntamb:
In addition to maintaining consistency between the API response and the Openapi spec, also refactored queries inside the `is_configured` api to only count those systems which have a performance profile similar to the query for `systems` api.

Also, did a minor refactoring to define common/default queries required by all APIs in one method.

EDIT - Removed the refactoring bit and would instead handle that change in a separate PR.

### DIFF
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -6,13 +6,11 @@ from flask_restful import Resource, abort, fields, marshal_with
 from ros.lib.models import (
     PerformanceProfile, RhAccount, System,
     db, RecommendationRating)
-from ros.lib.utils import (
-    is_valid_uuid, identity,
-    user_data_from_identity,
-    sort_io_dict, default_queries)
+from ros.lib.utils import is_valid_uuid, identity, user_data_from_identity, sort_io_dict
 from ros.api.common.pagination import (
     build_paginated_system_list_response,
-    limit_value, offset_value)
+    limit_value,
+    offset_value)
 import logging
 
 
@@ -29,26 +27,17 @@ SYSTEM_COLUMNS = [
 
 class IsROSConfiguredApi(Resource):
     def get(self):
-        account_number = identity(request)['identity']['account_number']
-        system_query = default_queries(account_number)
-
-        last_reported = (
-            db.session.query(PerformanceProfile.system_id, func.max(PerformanceProfile.report_date).label('max_date')
-                             )
-            .filter(PerformanceProfile.system_id.in_(system_query.subquery()))
-            .group_by(PerformanceProfile.system_id)
-            .subquery()
-        )
-        query = (
-            db.session.query(PerformanceProfile, System, RhAccount)
-            .join(last_reported, (last_reported.c.max_date == PerformanceProfile.report_date) &
-                  (PerformanceProfile.system_id == last_reported.c.system_id))
-            .join(System, System.id == last_reported.c.system_id)
-            .join(RhAccount, RhAccount.id == System.account_id)
-        )
-        system_count = query.count()
-        systems_with_suggestions = query.filter(System.number_of_recommendations > 0).count()
-        systems_waiting_for_data = query.filter(System.state == 'Waiting for data').count()
+        ident = identity(request)['identity']
+        account_query = db.session.query(RhAccount.id).filter(
+            RhAccount.account == ident['account_number']).subquery()
+        system_count = db.session.query(System.id)\
+            .filter(System.account_id.in_(account_query)).count()
+        systems_with_suggestions = db.session.query(System.id)\
+            .filter(System.account_id.in_(account_query))\
+            .filter(System.number_of_recommendations > 0).count()
+        systems_waiting_for_data = db.session.query(System.id)\
+            .filter(System.account_id.in_(account_query))\
+            .filter(System.state == 'Waiting for data').count()
 
         if system_count <= 0:
             status, code = False, 'NO_SYSTEMS'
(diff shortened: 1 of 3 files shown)


ros/api/v1/hosts.py
```diff
@@ -25,30 +25,46 @@
...
+        system_query = default_queries(ident)
 
+        last_reported = (
+            db.session.query(PerformanceProfile.system_id, func.max(PerformanceProfile.report_date).label('max_date')
+                             )
+            .filter(PerformanceProfile.system_id.in_(system_query.subquery()))
+            .group_by(PerformanceProfile.system_id)
+            .subquery()
+        )
+        print(last_reported)
```
saltgen, line 48: Shall we use LOG.info, in case we need this one?

ros/api/v1/hosts.py
```diff
@@ -25,30 +25,46 @@
...
-            }
+        system_query = default_queries(ident)
 
+        last_reported = (
+            db.session.query(PerformanceProfile.system_id, func.max(PerformanceProfile.report_date).label('max_date')
+                             )
+            .filter(PerformanceProfile.system_id.in_(system_query.subquery()))
+            .group_by(PerformanceProfile.system_id)
+            .subquery()
+        )
```
saltgen, line 47: :ok_hand:

ros/api/v1/hosts.py
apuntamb, line 48: @saltgen need to remove the print statement. Forgot to do so. :)

ros/api/v1/hosts.py
```diff
@@ -25,30 +25,46 @@
 ]
 
 
+def default_queries(ident, additional_filter=None):
+    account_query = db.session.query(RhAccount.id).filter(RhAccount.account == ident['account_number']).subquery()
+    if additional_filter is None:
+        return db.session.query(System.id).filter(System.account_id.in_(account_query))
+    return db.session.query(System.id).filter(System.account_id.in_(account_query)).filter(additional_filter)
+
```
saltgen, line 33: Thanks for the modularization :raised_hands: 
Looks like only the `account_number` is required from `ident`, shall we pass just the account_number
in such a case?
Also, I think we can move the function to utils.py or some other file that would be appropriate

ros/api/v1/hosts.py
saltgen: Np. I donno why I deleted the related comment :rofl: Oh well

saltgen:
Done with the comments for now. Thanks for the changes! :smiley:

ros/api/v1/hosts.py
apuntamb: @saltgen You mean directly passing `ident['account_number']` instead of `ident`? Because if so, in any case we have to fetch the account number from ident, so I don't foresee much of a difference.
Do let me know if that sounds okay or not?

PreetiW (approved):
@apuntamb Thank you for making the changes, I have tested this PR locally and it is working fine. Also checked it against Dashboard code and it won't break anything there

image

image

ros/api/v1/hosts.py
saltgen: Agreed, there is not much real-time gain in this, but in terms of readability, IMO this looks clearer.

```
def default_queries(account_number: str, additional_filter: bool = None):
    ...
    ...
```

P.S. Added the type hints just to convey what I really mean, not required to add them.

apuntamb: Done.

PreetiW (changes requested):
Hello @apuntamb 
I was checking filters functionality and the API is giving 500 due to attached screenshot
image

can you please check the reason behind this?

PreetiW (approved):
Thank you looking at the issue @apuntamb , checked with the latest commit and filtering is working fine. I also checked systems details page and it is working as ecpected 💯 

image

image

apuntamb:
@saltgen Any additional comments or can we merge this?

ros/api/v1/hosts.py
```diff
@@ -27,28 +29,34 @@
...
+        account_number = identity(request)['identity']['account_number']
+        system_query = default_queries(account_number)
 
-        if system_count <= 0:
-            return {
-                'success': False,
-                'code': 'NOSYSTEMS',
-                'count': system_count
-            }
+        last_reported = (
```
kgaikwad, line 35: I would like to understand why did you add this extra query?

ros/api/v1/hosts.py
```diff
@@ -27,28 +29,34 @@
...
+            db.session.query(PerformanceProfile.system_id, func.max(PerformanceProfile.report_date).label('max_date')
+                             )
+            .filter(PerformanceProfile.system_id.in_(system_query.subquery()))
+            .group_by(PerformanceProfile.system_id)
+            .subquery()
+        )
+        query = (
+            db.session.query(PerformanceProfile, System, RhAccount)
```
kgaikwad, line 43: Any specific reason you are loading all columns from all three tables and that is to get counts?

kgaikwad (changes requested):

saltgen:
> @saltgen Any additional comments or can we merge this?

Nope, lgtm. Thanks @apuntamb! :raised_hands:

ros/api/v1/hosts.py
apuntamb: @kgaikwad IMO, inside the /systems api, systems are listed based on Performance Profile filter i.e. we show systems only which have Performance Profile associated with the respective system.
It would make more sense to use the same query inside /is_configured api to keep the system counts consistent in both api responses.
Thoughts?

saltgen:
@apuntamb On second thought could you add an assert comparison in the test `test_is_configured`, between
count from systems endpoint and is_configured?

ros/api/v1/hosts.py
```diff
@@ -30,25 +30,32 @@ def get(self):
...
-        system_count = db.session.query(System.id)\
-            .filter(System.account_id.in_(account_query)).count()
-        systems_with_suggestions = db.session.query(System.id)\
-            .filter(System.account_id.in_(account_query))\
-            .filter(System.number_of_recommendations > 0).count()
-        systems_waiting_for_data = db.session.query(System.id)\
-            .filter(System.account_id.in_(account_query))\
-            .filter(System.state == 'Waiting for data').count()
+        system_query = db.session.query(System.id).filter(System.account_id.in_(account_query))
+        last_reported = (
```
kgaikwad, line 34: @apuntamb  - Could you please add new issue for checking PerformanceProfile records?
Original issue only talks about inconsistent response formats.

tests/test_api_endpoints.py
```diff
@@ -38,7 +38,7 @@ def test_status():
         assert response.json["status"] == "Application is running!"
 
 
-def test_is_configured(auth_token, db_setup, db_create_account, db_create_system):
+def test_is_configured(auth_token, db_setup, db_create_account, db_create_system, db_create_performance_profile):
```
kgaikwad, line 41: @apuntamb  - why `db_create_performance_profile` required here?

apuntamb: fixed.

apuntamb:
@kgaikwad As discussed over call with @pgarciaq , will handle **not** showing all systems i.e. **only** those systems which are shown through `/systems` api response in `is_configured` api too, in a separate PR.

kgaikwad (approved):
LGTM :+1: 
As discussed, refactoring and mismatched system count issue should be handled separately.

PR #156, opened 2021-12-23
RHIROS-449 Get region from performance_record

r14chandra:
Getting `region` value from rule hit won't work for optimized state. Not able to get `region` for NO_PCP_DATA also. Hence, using performance_record to get the `region` value, storing it in `systems` table and discarding it from `performance_record.` Also discarding `instance_type` from performance_record as we already have instance_type column in systems table.
Added tests for no-pcp, undersized and optimized states

### DIFF
--- a/ros/processor/insights_engine_result_consumer.py
+++ b/ros/processor/insights_engine_result_consumer.py
@@ -130,7 +130,7 @@ class InsightsEngineResultConsumer:
                     number_of_recommendations=rec_count,
                     state=SYSTEM_STATES[state_key],
                     instance_type=performance_record.get('instance_type'),
-                    region=reports[0].get('details').get('region')
+                    region=performance_record.get('region')
                 )
                 LOG.info(
                     f"{self.prefix} - System created/updated successfully: {host['id']}"
@@ -159,7 +159,8 @@ class InsightsEngineResultConsumer:
                         'max_io': 0,
                         'io': {}
                     }
-
+                del performance_record['instance_type']
+                del performance_record['region']
                 get_or_create(
                     db.session, PerformanceProfile, ['system_id', 'report_date'],
                     system_id=system.id,
--- a/ros/processor/insights_engine_result_consumer.py
+++ b/ros/processor/insights_engine_result_consumer.py
@@ -159,8 +159,10 @@ class InsightsEngineResultConsumer:
                         'max_io': 0,
                         'io': {}
                     }
+                # Following are saved on respective system record
                 del performance_record['instance_type']
                 del performance_record['region']
+
                 get_or_create(
                     db.session, PerformanceProfile, ['system_id', 'report_date'],
                     system_id=system.id,
--- a/tests/test_insights_engine_result_consumer.py
+++ b/tests/test_insights_engine_result_consumer.py
@@ -120,3 +120,19 @@ def test_process_report_undersized(engine_result_message, engine_consumer, db_se
         assert data.state == SYSTEM_STATES['INSTANCE_UNDERSIZED']
         assert db.session.query(PerformanceProfile).filter_by(system_id=data.id).first().performance_record ==\
                performance_record
+
+
+def test_process_report_optimized(engine_result_message, engine_consumer, db_setup, performance_record):
+    engine_result_message = engine_result_message("insights-engine-result-undersized.json")
+    host = engine_result_message["input"]["host"]
+    ros_reports = []
+    system_metadata = engine_result_message["results"]["system"]["metadata"]
+    current_performance_record = copy.copy(performance_record)
+    engine_consumer.process_report(host, ros_reports, system_metadata, performance_record)
+    data = db_get_host(host['id'])
+    assert str(data.inventory_id) == host['id']
+    with app.app_context():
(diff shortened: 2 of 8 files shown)


saltgen:
@r14chandra I will post my comments by 1st half tomo, sorry for the delay.
Meanwhile could you add in the tests for optimized, no-pcp and undersized?

tests/test_insights_engine_result_consumer.py
```diff
@@ -44,42 +47,92 @@ def engine_consumer():
...
+    with app.app_context():
+        assert data.instance_type == current_performance_record['instance_type']
+        assert data.state == SYSTEM_STATES['INSTANCE_UNDERSIZED']
+        assert db.session.query(PerformanceProfile).filter_by(system_id=data.id).first().performance_record ==\
+               performance_record
+
+
+def test_process_report_optimized(engine_result_message, engine_consumer, db_setup, performance_record):
+    engine_result_message = engine_result_message("insights-engine-result-undersized.json")
+    host = engine_result_message["input"]["host"]
```
saltgen, line 127: @r14chandra Line 126 should load engine msg for optmized i.e. insights-engine-result-`optimized`.json

r14chandra: Typo 😓. Thanks for pointing out 👍🏻

ros/processor/insights_engine_result_consumer.py
```diff
@@ -130,7 +130,7 @@ def process_report(self, host, reports, utilization_info, performance_record):
                     number_of_recommendations=rec_count,
                     state=SYSTEM_STATES[state_key],
                     instance_type=performance_record.get('instance_type'),
-                    region=reports[0].get('details').get('region')
+                    region=performance_record.get('region')
                 )
```
saltgen, line 134: The idea of adding tests generally is to ensure that your changes are actually working.
Could you assert the `region` data in the system state tests i.e. along with the `instance_type`?

tests/test_insights_engine_result_consumer.py
```diff
@@ -44,42 +47,92 @@ def engine_consumer():
...
+               performance_record
+
+
+def test_process_report_undersized(engine_result_message, engine_consumer, db_setup, performance_record):
+    engine_result_message = engine_result_message("insights-engine-result-undersized.json")
+    host = engine_result_message["input"]["host"]
+    ros_reports = [engine_result_message["results"]["reports"][7]]
+    system_metadata = engine_result_message["results"]["system"]["metadata"]
+    current_performance_record = copy.copy(performance_record)
+    engine_consumer.process_report(host, ros_reports, system_metadata, performance_record)
```
saltgen, line 115: Let's rename `current_performance_record` to `_performance_record` or something more appropriate :thinking:

saltgen:
@r14chandra I tried loading `insights-engine-result-under-pressure.json` in `test_process_report_optimized`
And it is passing :open_mouth: 
Can you check what's up with that?

tests/test_insights_engine_result_consumer.py
saltgen: Np!

r14chandra:
> @r14chandra I tried loading `insights-engine-result-under-pressure.json` in `test_process_report_optimized` And it is passing open_mouth Can you check what's up with that?

@saltgen I guess we're utilizing the insights-engine-result for this particular state (i.e Optimized) for getting  `system_metadata`  and `host` and not `ros_report` (which is unique for every state). That's why any engine would work for this state. Please, correct me.

tests/test_insights_engine_result_consumer.py
r14chandra: Yeah `_performance_record` would be fine 👍🏻

ros/processor/insights_engine_result_consumer.py
r14chandra: `instance_type` assertion is already there. I assert the region data 👍🏻

ros/processor/insights_engine_result_consumer.py
```diff
@@ -159,7 +159,8 @@ def process_report(self, host, reports, utilization_info, performance_record):
                         'max_io': 0,
                         'io': {}
                     }
-
+                del performance_record['instance_type']
+                del performance_record['region']
                 get_or_create(
```
saltgen, line 164: @r14chandra Could you add a comment above, stating the reason for the `del` statements?

saltgen:
> > @r14chandra I tried loading `insights-engine-result-under-pressure.json` in `test_process_report_optimized` And it is passing open_mouth Can you check what's up with that?
> 
> @saltgen I guess we're utilizing the insights-engine-result for this particular state (i.e Optimized) for getting `system_metadata` and `host` and not `ros_report` (which is unique for every state). That's why any engine result would work for this state. Please, correct me.

This was primarily due to ros_reports being explicitly set to [] in the respective test, which should be the case but only for optimized state. All good, thanks!

saltgen (approved):
Appreciate the effort on the tests @r14chandra  :raised_hands: 
LGTM :+1:

kgaikwad:
Thank you @saltgen for review! Merging this PR.

PR #157, opened 2021-12-23
RHIROS-425 | unittests for API endpoints

saltgen:
### Changes:

1. Added tests for all endpoints
2. Removed an un-necessary variable
3. Added a test archive ~ optimized


apuntamb (approved):
Looks clean. Nice work @saltgen !

ros/api/v1/recommendation_ratings.py
```diff
@@ -47,7 +47,6 @@ def post(self):
             RecommendationRating.system_id == system.id,
             RecommendationRating.rated_by == username).first()
 
-        status_code = None
```
apuntamb, line 50: @saltgen Just one question though, if you remove this initialization from here, will this work https://github.com/RedHatInsights/ros-backend/blob/main/ros/api/v1/recommendation_ratings.py#L66 ?

saltgen: Yep, pretty sure it will, cause the flow will go through `if and else` Line #51 onwards no matter what, and then status_code will get initialized :+1:

apuntamb:
Merged, thanks @saltgen !!

PR #158, opened 2021-12-24
Ref-RHIROS-424 set 'ENABLE_RBAC' to True 

r14chandra:


saltgen:
@r14chandra This may not be necessary, I will update in a day or so.

PR #161, opened 2022-01-05
RHIROS-457 Fix system counts of is_configured & systems api and refactor

apuntamb:


tests/test_api_endpoints.py
```diff
@@ -38,7 +38,7 @@ def test_status():
         assert response.json["status"] == "Application is running!"
 
 
-def test_is_configured(auth_token, db_setup, db_create_account, db_create_system):
+def test_is_configured(auth_token, db_setup, db_create_account, db_create_system, db_create_performance_profile):
     with app.test_client() as client:
         response = client.get('/api/ros/v1/is_configured', headers={"x-rh-identity": auth_token})
         assert response.status_code == 200
```
saltgen, line 44: @apuntamb Could you add an assert comparison in the `test_is_configured`, between count from systems endpoint and is_configured?
Also I think we can move `test_systems` above `test_is_configured`, since we will need systems test to run fine before pytest moves on to `test_is_configured`

saltgen (commented):
Checked against filters on this [one](https://github.com/RedHatInsights/ros-backend/pull/155#pullrequestreview-839882090)  and multiple systems, @apuntamb I don't see any logical issues :+1:

ros/api/v1/hosts.py
```diff
@@ -27,17 +30,25 @@
...
-            .filter(System.state == 'Waiting for data').count()
+        account_number = identity(request)['identity']['account_number']
+        system_query = default_queries(account_number)
+        last_reported = (
+            db.session.query(PerformanceProfile.system_id, func.max(PerformanceProfile.report_date).label('max_date')
+                             )
+            .filter(PerformanceProfile.system_id.in_(system_query.subquery()))
+            .group_by(PerformanceProfile.system_id)
+            .subquery()
+        )
```
kgaikwad, line 41: Instead of making a use aggregate function to get last reported, can't we simply use a query to select distinct system ids which has  `PerformanceProfile` records?

ros/api/v1/hosts.py
```diff
@@ -27,17 +30,25 @@
...
+        query = (
+            db.session.query(PerformanceProfile.system_id, System.id, RhAccount.id)
```
kgaikwad, line 43: In my opinion, `System.id` will be sufficient to get systems count. Any thoughts?

ros/api/v1/hosts.py
```diff
@@ -27,17 +30,25 @@
...
+            .join(last_reported, (last_reported.c.max_date == PerformanceProfile.report_date) &
+                  (PerformanceProfile.system_id == last_reported.c.system_id))
+            .join(System, System.id == last_reported.c.system_id)
+            .join(RhAccount, RhAccount.id == System.account_id)
```
kgaikwad, line 47: I can see you applied a default query in above query then why did you require to join `RhAccount` table?

kgaikwad (changes requested):
@apuntamb - thank you for fixing the counts mismatch!
Added few inline comments.

ros/api/v1/hosts.py
```diff
@@ -27,17 +30,15 @@
...
-        systems_waiting_for_data = db.session.query(System.id)\
-            .filter(System.account_id.in_(account_query))\
-            .filter(System.state == 'Waiting for data').count()
+        query = (
+            db.session.query(System.id)
+            .join(PerformanceProfile, PerformanceProfile.system_id == System.id)
+            .join(RhAccount, RhAccount.id == System.account_id)
+            .distinct()
+        )
+        system_count = query.count()
```
kgaikwad, line 39: @apuntamb  - It should only show systems count specific to current user's account.With your changes it will show all systems count to every user regardless of her/his account.

apuntamb: Yeah, missed that. Made certain changes for the same.

ros/api/v1/hosts.py
```diff
@@ -27,17 +30,16 @@
...
-        systems_waiting_for_data = db.session.query(System.id)\
-            .filter(System.account_id.in_(account_query))\
-            .filter(System.state == 'Waiting for data').count()
+        account_number = identity(request)['identity']['account_number']
+        system_query = default_queries(account_number)
+        query = (
+            db.session.query(System.id)
+            .filter(PerformanceProfile.system_id.in_(system_query.subquery()))
+            .distinct()
+        )
```
kgaikwad, line 39: @apuntamb  - Is this query working correctly on local?

apuntamb: yes @kgaikwad

ros/lib/utils.py
```diff
@@ -105,3 +106,8 @@ def sort_io_dict(performance_utilization: dict):
     performance_utilization.update({**sorted_io_dict})
     del performance_utilization['io']
     return performance_utilization
+
+
+def default_queries(account_number):
```
kgaikwad, line 111:
```suggestion
def system_ids_by_account(account_number):
```
In my opinion, as it handles account filter i.e main scope it better to rename it to something specific.

r14chandra:
@apuntamb @kgaikwad  I've tested it locally. For a system with multiple PP, we're getting single count for that. Also, the count is now showing number of systems with PP associated with a particular account.

r14chandra (approved):

kgaikwad (approved):
Codewise looks legit

PR #162, opened 2022-01-06
RHIROS-460 | Util value types on openapi.json

saltgen:
Changes:
1. cpu and memory types on SystemDetails schema and SystemItem schema

Reasons:
1. Value types are referred by server on deserialization


kgaikwad (approved):
LGTM

PR #164, opened 2022-01-07
RHIROS-211 | Operating system details for respective host/system

saltgen:
### Changes:
1. Added release_version field on model
2. Added migration file
3. Initialized system with release value on system record creation
4. Added relevant test
5. Changes for availability of release information on _systems_ and _systems detail_ endpoints, introduced test coverage


r14chandra:
I've one concern. Do we want to show `rhel_version` or `release`. Referring the jira card, it says which version of RHEL. Any thoughts?

saltgen:
> I've one concern. Do we want to show `rhel_version` or `release`? Referring the jira card, it says which version of RHEL. Any thoughts?

The official term is _release version_, [here](https://access.redhat.com/discussions/1312334).
Open to suggestions though

r14chandra:
> > I've one concern. Do we want to show `rhel_version` or `release`? Referring the jira card, it says which version of RHEL. Any thoughts?
> 
> The general term is _release version_, [here](https://access.redhat.com/discussions/1312334). Open to suggestions though

I mean `"rhel_version": "8.4"` or `"release": "Red Hat Enterprise Linux release 8.4 (Ootpa)"`

kgaikwad:
@saltgen, could you please rebase and update changes?

migrations/versions/d79024110f49_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'd79024110f49'
+down_revision = 'db7f5df23e1b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('systems', sa.Column('release_version', sa.String(length=100), nullable=True))
```
kgaikwad, line 21:
```suggestion
import sqlalchemy as sa
from sqlalchemy.dialects.postgresql import JSONB

# revision identifiers, used by Alembic.
revision = 'd79024110f49'
down_revision = 'db7f5df23e1b'
branch_labels = None
depends_on = None


def upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    op.add_column('systems', sa.Column('operating_system',  postgresql.JSONB(astext_type=sa.Text()), nullable=True))
```

For reference - https://github.com/RedHatInsights/inventory-schemas/blob/master/schemas/system_profile/v1.yaml#L252

migrations/versions/d79024110f49_.py
```diff
@@ -0,0 +1,28 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('systems', 'release_version')
```
kgaikwad, line 27:
```suggestion
    op.drop_column('systems', 'operating_system')
```

ditto

kgaikwad (changes requested):
> @kgaikwad Do you concur?

Inventory kafka message i.e `platform.inventory.events` includes os information under `system_profile > operating_system` - {'name': 'RHEL', 'major': 8, 'minor': 4} 
In my opinion, will it be good idea to store it into `operating_system` field ?  In this way from UI point of view as well, we will be consistent with other services. 
Example -
Screenshot from 2022-02-09 18-56-15

What do you think?

saltgen:
@kgaikwad In addition to adding `operating_system` in events consumer, IIUC we will need to add the same in the
engine consumer as well else the system will have the value only when it receives event msgs.
Shall I proceed?

kgaikwad:
> @kgaikwad In addition to adding `operating_system` in events consumer, IIUC we will need to add the same in the engine consumer as well else the system will have the value only when it receives event msgs. Shall I proceed?

In my opinion, adding/updating os info from events message will serve the purpose. Any scenario that am I missing here?

migrations/versions/d79024110f49_.py
saltgen: Changes :rocket:

migrations/versions/d79024110f49_.py
saltgen: Replace migration file with new one. Changes :rocket:!

saltgen:
> 



> > @kgaikwad In addition to adding `operating_system` in events consumer, IIUC we will need to add the same in the engine consumer as well else the system will have the value only when it receives event msgs. Shall I proceed?
> 
> In my opinion, adding/updating os info from events message will serve the purpose. Any scenario that am I missing here?

@kgaikwad I will revisit as to why exactly we have a `get_or_create` system in the engine consumer as well, just to be sure.
Atm I believe it is only to update an already created system.

Considering the suggestion I have :rocket: all  the changes, thanks!

PreetiW (approved):
Thanks @saltgen, I tested these changes against the frontend and it is working as expected.
The sorting was giving error but since we have seperate cards for handling sorting and filtering IMO we can merge this one.

recording link: https://drive.google.com/file/d/1za7tB2XV0UfxW2VeZRedeI7dDF5Ii8Lk/view?usp=sharing

saltgen:
> Thanks @saltgen, I tested these changes against the frontend and it is working as expected. The sorting was giving error but since we have seperate cards for handling sorting and filtering IMO we can merge this one.
> 
> recording link: https://drive.google.com/file/d/1za7tB2XV0UfxW2VeZRedeI7dDF5Ii8Lk/view?usp=sharing

@PreetiW I would like to get sorting done on this MR, as backend changes would be very minimal, tests are also in place now.
You may add necessary changes for sorting on a new PR later for fe.

@kgaikwad For sorting we will need to save the `major.minor` version values as floats instead of `str`, shall I proceed with the changes?

PreetiW:
@saltgen the front-end changes are already in place for sorting, once the backend changes are done let me know and I will test it against the FE.

kgaikwad (changes requested):
+1 for new PR for sorting instead of blocking this PR.

kgaikwad:
@PreetiW  - For better user experience, could you disable sorting in OS column on frontend? After backend change i.e OS sorting, it would be good to enable sorting on the OS column.

PreetiW:
Sure @kgaikwad I agree, I will check the front-end side on how to disable sorting for specific column and will do that.

migrations/versions/605a688d0cd2_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '605a688d0cd2'
+down_revision = 'db7f5df23e1b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('systems', sa.Column('operating_system', sa.String(length=25), nullable=True))
```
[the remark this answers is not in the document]
saltgen: @kgaikwad Change pushed, field on db looks like this now, response on API looks the same.
`operating_system          | {"name": "RHEL", "major": 8, "minor": 4}`
Let me know of this needs changes :arrow_up:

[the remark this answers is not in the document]
kgaikwad: @saltgen - Any thoughts on how frontend should consume and use it? OR Will it be OK to add new readonly property which makes easy for frontend to display values per row?

@PreetiW  ^^ what do you think?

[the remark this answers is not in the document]
saltgen: @kgaikwad API response value remains unchanged, `"operating_system": "RHEL 8.4"`

saltgen:
@kgaikwad As discussed on Slack, API key is now `os` instead of `operating_system`
cc: @PreetiW 

Changes :rocket:

ros/openapi/openapi.json
```diff
@@ -365,6 +365,9 @@
                     },
                     "idling_time": {
                         "type": "number"
+                    },
+                    "operating_system": {
```
kgaikwad, line 369:
```suggestion
                    "os": {
```

kgaikwad (changes requested):

ros/openapi/openapi.json
saltgen: @kgaikwad my bad :sweat_smile: , change pushed

ros/openapi/openapi.json
```diff
@@ -428,6 +431,9 @@
...
```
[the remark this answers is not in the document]
saltgen: Change pushed

kgaikwad (approved):

kgaikwad:
Thanks @saltgen and @PreetiW for review!

PR #166, opened 2022-01-10
RHIROS-348 - add grafana dashboard

kgaikwad:


apuntamb (approved):
Since @kgaikwad you did demo grafana dashboard earlier in working state, the PR looks good to me in that case.
@saltgen Do you have any suggestions or comments?

kgaikwad:
@apuntamb - Yes, I think we should merge this PR so that I can test stage cycle end to end for dashboard change.

PR #170, opened 2022-01-11
RHIROS-469 | (Cherry-picked) orderByParam is now max_io instead of io, systems endpoint

saltgen:
Changes:
1. Updated orderByParam from io to max_io

Reasons:
1. io is now longer a sortable parameter


apuntamb (approved):
lgtm!

apuntamb:
merged, thanks @saltgen

PR #171, opened 2022-01-12
RHIROS-475 | OpenAPI docs misc. ROS v1

saltgen:
Changes:
1. Ordered util values as they appear on API response
2. Modified limit param description text
3. Added inventory_id on PerformanceHistory schema
4. Updated host_id to inventory_id in SuggestionsList schema
5. Removed un-available state filter parameters

Reasons:
* Consistency

### DIFF
--- a/ros/openapi/openapi.json
+++ b/ros/openapi/openapi.json
@@ -343,10 +343,10 @@
                     "performance_utilization": {
                         "type": "object",
                         "properties": {
-                            "memory": {
+                            "cpu": {
                                 "type": "integer"
                             },
-                            "cpu": {
+                            "memory": {
                                 "type": "integer"
                             },
                             "max_io": {
@@ -393,10 +393,10 @@
                     "performance_utilization": {
                         "type": "object",
                         "properties": {
-                            "memory": {
+                            "cpu": {
                                 "type": "integer"
                             },
-                            "cpu": {
+                            "memory": {
                                 "type": "integer"
                             },
                             "max_io": {
@@ -440,6 +440,9 @@
                     "links": {
                         "$ref": "#/components/schemas/Links"
                     },
+                    "inventory_id": {
+                        "type": "string"
+                    },
                     "data": {
                         "type": "array",
                         "items": {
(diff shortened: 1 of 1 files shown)


ros/openapi/openapi.json
```diff
@@ -537,7 +540,7 @@
                 "name": "limit",
                 "in": "query",
                 "required": false,
-                "description": "Number of items to return, set -1 to return all",
+                "description": "Number of items to return",
```
saltgen, line 543: @kgaikwad I hope this is okay for now, as keeping that text would mean we will need code changes.
-1 is not returning systems at the moment.

ros/openapi/openapi.json
```diff
@@ -605,11 +608,9 @@
                         "type": "string",
                         "enum": [
                             "Idling",
-                            "Oversized",
                             "Undersized",
                             "Optimized",
                             "Under pressure",
-                            "Storage rightsizing",
                             "Waiting for data"
                         ]
```
saltgen, line 615: @kgaikwad Shall I remove "Undersized" as well?

kgaikwad: In my opinion, we can keep it as it is.

ros/openapi/openapi.json
```diff
@@ -605,11 +608,9 @@
                         "type": "string",
                         "enum": [
                             "Idling",
-                            "Oversized",
```
r14chandra, line 608: @saltgen Why are we removing `Oversized` state? We're supporting this state currently (But not tested though)

saltgen: @r14chandra AFAIK we don't have an Oversized state atm, for that we will need integration with sources.
So IMO since we don't support it yet there is no need to have it in the docs for now

kgaikwad: @saltgen - if you remove it here then frontend also needs to be updated. In my opinion let's keep this one.

r14chandra: +1 for keeping `oversized` state

saltgen: @kgaikwad Ah yes, makes sense, change pushed.

r14chandra (approved):
Looks great 👍🏻 @saltgen. Tested locally 💯

ros/openapi/openapi.json
kgaikwad: Make sense!

kgaikwad (approved):
Codewise looks good!

PR #178, opened 2022-02-01
Refs RHIROS #133 -  Fix rbac configs for clowder

apuntamb:


apuntamb:
@kgaikwad it is because of the current pr_check job. Hopefully, it should be fixed in #179 . Once that is merged, I can rebase this one.

kgaikwad:
@apuntamb @mayurilahane  - From above comment it seems that problem is not due to PR change and @apuntamb this is different job than one which runs on main branch merge. So merging this change.

PR #180, opened 2022-02-03
add testing plugin in deployment yaml

digitronik:


digitronik:
@kgaikwad @apuntamb I think we can merge this now. This change only tells deployment that uses testing plugin ros.

After this MR in I can check for #182  run.

clowdapp.yaml
```diff
@@ -83,6 +83,8 @@ objects:
         partitions: 1
       - topicName: platform.engine.results
         partitions: 1
+    testing:
+      iqePlugin: ros
```
apuntamb, line 87: @digitronik Can you remove the following lines: https://github.com/RedHatInsights/ros-backend/pull/180/files#diff-0f518b5985b4ea86412d39c9757171411345db257cb061b84bb314a7a2b3d5e9R15-R17
and instead add `- engine` below `- rbac` here: https://github.com/RedHatInsights/ros-backend/pull/180/files#diff-0f518b5985b4ea86412d39c9757171411345db257cb061b84bb314a7a2b3d5e9R14

apuntamb: cc @r14chandra

digitronik: gitlab connection failing.

r14chandra: @apuntamb Is it because it's a private repo (for insights-engine)?

digitronik: @r14chandra I think yes. As we are using ext Jenkin this repo maybe not be accessible.

r14chandra: Yes @digitronik. So, we need to use ci-int. Is there any thing to be considered for ci-int/ci-ext ?  @apuntamb

apuntamb: @r14chandra @digitronik ! IMHO, we should first check if there is a way we can communicate with the internal repo using ci-ext, since it is a dependency.

apuntamb: @r14chandra and @digitronik As you mentioned, we need to move to ci-int.
To consider the same, we need to update the webhook in github repo as well as update the job to ci-int in app-interface. Let me fix these two pre-requisites and check if the build is succeeding for th app-interface MR, and after that is merged, we can retest this PR.

apuntamb (approved):
Approving this so that @digitronik can test the current deployment on appsre and next steps to investigate the ci build failures could then follow.

apuntamb:
Merged @digitronik , thanks!

PR #182, opened 2022-02-14
Renaming Clowder deployments

saltgen:
This is w.r.t to  rollback on prod.

Issue: Traffic on prod is currently being routed to non-existent clowder pods(these don't spin up for some reason)

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -16,7 +16,7 @@ objects:
     - puptoo
     - ingress
     deployments:
-    - name: backend-service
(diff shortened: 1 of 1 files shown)


kgaikwad (approved):
Naming wise changes looks good to me. 

Not particularly to this PR change but one thing I noticed that with old deployment templates, ROS has two services whereas with new clowder changes it is having only one service. In case need any additional change after verifying the scenario, I would be good to address it in separate PR.

PR #184, opened 2022-02-24
Fix seeding in clowdapp

apuntamb:


saltgen:
I am not sure about the order in which the deployments spin up, hence IMO we can add db upgrade for the processor as well, [here](https://github.com/RedHatInsights/ros-backend/blob/5849d30976e70dbc2d6247e0b404a14c054444ec/clowdapp.yaml#L65). 

I have encountered issues locally when I try to power up the processor without running migrations.

Alternatively can we run the migrations and seeding [here](https://github.com/RedHatInsights/ros-backend/blob/5849d30976e70dbc2d6247e0b404a14c054444ec/clowdapp.yaml#L76)? IMO worth a shot :slightly_smiling_face:

apuntamb:
@saltgen In case of local devel setup, there is a scope of running the `processor` independently, but for environments like stage or prod, we do not expect processor to run without api or db components.
Also, I considered following references to understand better about this and looks like it might not be a best practice to run migrations in `database` specification:

1. [vulnerability-engine](https://github.com/RedHatInsights/vulnerability-engine/blob/master/deploy/clowdapp.yaml#L594)
2. [compliance-backend](https://github.com/RedHatInsights/compliance-backend/blob/master/clowdapp.yaml#L20)
3. [patchman-engine](https://github.com/RedHatInsights/patchman-engine/blob/master/deploy/clowdapp.yaml#L327)

What do you think?

saltgen:
@apuntamb I see in that case I think we are good to go with this one :+1:

saltgen (approved):

apuntamb:
Thanks @saltgen for the review, merging!

PR #163, opened 2022-01-07
RHIROS-391 | Log streaming to CloudWatch -> Kibana Dashboard

saltgen:
### Changes:
1. Added cw_logging.py module
2. Introduced relevant environment variables vars on config.py
3. Added CloudWatch streaming function calls in processor and api
4. Added dependencies on Pipfile and regenerated Pipfile.lock
5. Cleaned up a pesky print statement from test_system_detail

### Reasons:
1. Houses streaming handler function
2. Necessary for authenticating with relevant AWS account
3. Invokes the handler function
4. Mentioning and locking dependencies

### Additional:

* Tests have not been added since 95% of the changes is meant to function on Openshift

### References:

- [Documentation](https://consoledot.pages.redhat.com/docs/dev/platform-documentation/tools/logging.html)
- [Sample code](https://github.com/RedHatInsights/vmaas/blob/master/vmaas/common/logging_utils.py)

PR #186, opened 2022-02-25
RHIROS-391 | Environment Variable for Cloudwatch Logging

saltgen:
### Changes:

1. Added inits for the var, clowder and non-clowder
2. Based on the same the AWS secret values are initialized
3. Added init of the variable on respective saas-template files as well

### Additional:
Logical reference, [here](https://github.com/RedHatInsights/patchman-engine/blob/9c4478fd30572877464f7bea69d5367ec4b11be4/base/utils/config.go#L105)
Related to #163

### DIFF
--- a/ros/lib/config.py
+++ b/ros/lib/config.py
@@ -41,6 +41,8 @@ if CLOWDER_ENABLED:
         if endpoint.app == "rbac":
             RBAC_SVC_URL = f"http://{endpoint.hostname}:{endpoint.port}"
             break
+    CW_ENABLED = True if LoadedConfig.Logging.Cloudwatch else False  # CloudWatch/Kibana Logging
+
 else:
     DB_NAME = os.getenv("ROS_DB_NAME", "postgres")
     DB_USER = os.getenv("ROS_DB_USER", "postgres")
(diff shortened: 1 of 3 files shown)


ros/lib/config.py
```diff
@@ -69,3 +72,10 @@ def get_logger(name):
 # Number of days after which data is considered to be outdated.
 DAYS_UNTIL_STALE = os.getenv("DAYS_UNTIL_STALE", 14)
 INSTANCE_PRICE_UNIT = 'USD/hour'
+
+if CW_ENABLED is True:
+    # Available only in k8s namespace, through an app-interface automation
+    AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID", None)
+    AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY", None)
+    AWS_REGION_NAME = os.getenv("AWS_REGION_NAME", None)
+    AWS_LOG_GROUP = os.getenv("AWS_LOG_GROUP", None)
```
apuntamb, line 81: Aren't these veriables required for Clowder Enabled state too?

saltgen: @apuntamb Yes, in clowder if else block I'm setting up the CW_ENABLED flag and then finally based on the flag I'm init-ing these values

r14chandra: @saltgen You mean these variables should already be set (we don't have to set the explicitly) if `CW_ENABLED = True` else default value will be `None`

r14chandra: `AWS_ACCESS_KEY_ID=app_common_python.LoadedConfig.logging.cloudwatch.accessKeyId`
Like this one ^^

saltgen: @apuntamb @r14chandra Thanks for the tip(clowder-native config values) :raised_hands: 
Changes pushed :rocket:

r14chandra (approved):
Looks great 👍🏻 @saltgen 🙌🏻

saltgen:
@apuntamb @r14chandra FYI, Env var changes have been moved to #163 as per suggestion

apuntamb:
@saltgen Okay sure! In that case, please close this PR in favor of #163 .

kgaikwad (approved):

kgaikwad:
As the changes are working fine on stage as per slack conversation. Removing `do not merge` label from PR.

PR #191, opened 2022-03-03
RHIROS-512  ROS CTA

r14chandra:
Payload for no. of systems greater than zero
Screenshot from 2022-03-03 15-58-36

Payload for no. of systems equals to zero 
Screenshot from 2022-03-03 15-12-19

### DIFF
--- a/ros/api/v1/call_to_action.py
+++ b/ros/api/v1/call_to_action.py
@@ -1,0 +1,59 @@
+from flask_restful import Resource
+from flask import request
+from ros.lib.utils import (
+    identity, system_ids_by_account)
+from ros.lib.models import (
+    System, PerformanceProfile, db)
+
+
+class CallToActionApi(Resource):
+    def get(self):
+        account_number = identity(request)['identity']['account_number']
+        system_query = system_ids_by_account(account_number)
+        query = (
+            db.session.query(PerformanceProfile.system_id)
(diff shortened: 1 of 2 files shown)


ros/api/v1/call_to_action.py
```diff
@@ -0,0 +1,59 @@
...
+                    "redhatInsights": [
+                        {
+                            "id": "ros-1",
+                            "description": "Resource optimization recommends to assess and monitor"
+                            + " cloud usage and optimization on these systems",
+                            "icon": "cog",
+                            "action": {
+                                "title": f"View {total_system_count} {suffix}"
+                                + " with suggestions",
+                                "href": "/insights/ros"
```
kgaikwad, line 53:
```suggestion
                                "href": "/insights/ros?with_suggestions=true"
```
As in title, it talks about how many systems with recommentations then will it be good idea to redirect to page where it displays the number of systems with suggestions?

kgaikwad (changes requested):
Thank you @r14chandra! 
Overall looks good! Added small inline comment.

ros/api/v1/call_to_action.py
r14chandra: Thanks @kgaikwad. Pushed the change 👍🏻

kgaikwad (approved):
LGTM :+1:

PR #193, opened 2022-03-04
Ref-RHIROS 512 fix query for total no. of systems

r14chandra:


kgaikwad (approved):

kgaikwad:
Tested! Working as expected.  Thanks @r14chandra

PR #194, opened 2022-03-07
RHIROS-565 | Handle records with no operating_system data

saltgen:
Changes:

1. Added conditional init for 'os' in system detail and systems api
2. Added relevant test for system detail endpoint

### DIFF
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -148,9 +148,15 @@ class HostsApi(Resource):
                     row.PerformanceProfile.performance_utilization
                 )
                 host['idling_time'] = row.PerformanceProfile.idling_time
-                host['os'] = f"{row.System.operating_system['name']} " \
-                             f"{row.System.operating_system['major']}." \
-                             f"{row.System.operating_system['minor']}"
+                if (
+                        row.System.operating_system
+                        and all(key in row.System.operating_system.keys() for key in ['name', 'major', 'minor'])
+                ):
+                    host['os'] = f"{row.System.operating_system['name']} " \
+                                 f"{row.System.operating_system['major']}." \
+                                 f"{row.System.operating_system['minor']}"
+                else:
+                    host['os'] = "N/A"
                 hosts.append(host)
             except Exception as err:
                 LOG.error(
@@ -275,9 +281,15 @@ class HostDetailsApi(Resource):
             record['rating'] = rating_record.rating if rating_record else None
             record['report_date'] = profile.report_date
             record['idling_time'] = profile.idling_time
-            record['os'] = f"{system.operating_system['name']} " \
-                           f"{system.operating_system['major']}." \
-                           f"{system.operating_system['minor']}"
+            if(
+                    system.operating_system
+                    and all(key in system.operating_system.keys() for key in ['name', 'major', 'minor'])
+            ):
+                record['os'] = f"{system.operating_system['name']} " \
+                               f"{system.operating_system['major']}." \
+                               f"{system.operating_system['minor']}"
+            else:
+                record['os'] = "N/A"
         else:
             abort(404, message="System {} doesn't exist"
                   .format(host_id))
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -148,14 +148,7 @@ class HostsApi(Resource):
                     row.PerformanceProfile.performance_utilization
                 )
                 host['idling_time'] = row.PerformanceProfile.idling_time
-                if (
-                        row.System.operating_system
-                        and all(key in row.System.operating_system.keys() for key in ['name', 'major', 'minor'])
-                ):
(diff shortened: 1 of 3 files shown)


PreetiW:
@saltgen on frontend I have a check that if the value is null it will show N/A so in case other fields in the backend sending null as values (IIRC provider, instance type returns null in case of waiting for data state) then you can update this to send null for consistency. Let me know your thoughts!

apuntamb:
@saltgen I would suggest the same to represent empty string or none value similar to how it is done for say, cloud_provider field and frontend can recognize and set it to N/A at it's end.

ros/api/v1/hosts.py
```diff
@@ -148,9 +148,15 @@ def get(self):
...
-                host['os'] = f"{row.System.operating_system['name']} " \
-                             f"{row.System.operating_system['major']}." \
-                             f"{row.System.operating_system['minor']}"
+                if (
+                        row.System.operating_system
+                        and all(key in row.System.operating_system.keys() for key in ['name', 'major', 'minor'])
+                ):
+                    host['os'] = f"{row.System.operating_system['name']} " \
+                                 f"{row.System.operating_system['major']}." \
+                                 f"{row.System.operating_system['minor']}"
```
apuntamb, line 157: How about we instead use `row.System.operating_system.get('name')` ?

saltgen:
> @saltgen I would suggest the same to represent empty string or none value similar to how it is done for say, cloud_provider field and frontend can recognize and set it to N/A at it's end.

@apuntamb Okay, lets stay consistent, I have also removed respective else blocks and added another test.
Change pushed!

ros/api/v1/hosts.py
```diff
@@ -148,9 +148,14 @@ def get(self):
...
-                             f"{row.System.operating_system['major']}." \
-                             f"{row.System.operating_system['minor']}"
+                if (
+                        row.System.operating_system
+                        and all(key in row.System.operating_system.keys() for key in ['name', 'major', 'minor'])
+                ):
+                    host['os'] = f"{row.System.operating_system.get('name')} " \
+                                 f"{row.System.operating_system.get('major')}." \
+                                 f"{row.System.operating_system.get('minor')}"
+
```
kgaikwad, line 158: It seems that this logic will be required more than one place. Will it be good idea to encapsulate into new readonly property on System class? What do you think?

saltgen: @kgaikwad Changes :rocket:

kgaikwad (approved):
Legit

r14chandra (approved):

PreetiW (commented):
Left a small query on the PR.

PreetiW:
@kgaikwad @saltgen sorry somehow the query got lost, my question was when os_data = None what will be the API response, will it be null or empty string for OS value.

PreetiW:
@kgaikwad @saltgen sorry somehow the query got lost, my question was when os_data = None what will be the API response, will it be null or empty string for OS value.

kgaikwad:
@PreetiW -

```
{
    "meta": {
        "count": 2,
        "limit": 10,
        "offset": 0
    },
    "links": {
        "first": "/api/ros/v1/systems?limit=10&offset=0",
        "last": "/api/ros/v1/systems?limit=10&offset=0",
        "next": null,
        "previous": null
    },
    "data": [
        {
            "fqdn": "ip-172-31-28-69.ec2.internal",
            "display_name": "ip-172-31-28-69.ec2.internal",
            "inventory_id": "2e966f13-e300-4024-9f4a-dd2219a40efd",
            "account": "6089719",
            "number_of_suggestions": 1,
            "state": "Under pressure",
            "performance_utilization": {
                "cpu": 10,
                "memory": 12,
                "max_io": 0.314,
                "io_all": {
                    "xvda": 0.314
                }
            },
            "cloud_provider": "aws",
            "instance_type": "t2.micro",
            "idling_time": "179.70",
            "os": "RHEL 8.4"
        },
        {
            "fqdn": "iqe-ros-a84cec78-1abf-40b4-bbbf-9ddd183786bb",
            "display_name": "iqe-ros-a84cec78-1abf-40b4-bbbf-9ddd183786bb",
            "inventory_id": "8f6c2b49-0136-450a-b103-95dc6404662e",
            "account": "6089719",
            "number_of_suggestions": 1,
            "state": "Idling",
            "performance_utilization": {
                "cpu": 0,
                "memory": 0,
                "max_io": 0.314,
                "io_all": {
                    "xvda": 0.314
                }
            },
            "cloud_provider": "aws",
            "instance_type": "t2.micro",
            "idling_time": "199.70",
            "os": null
        }
    ]
}
```

PreetiW:
Thanks for sharing @kgaikwad, no more queries from my side. 👍🏻

PreetiW (approved):

kgaikwad:
Thank you @saltgen!

PR #199, opened 2022-03-11
RHIROS-246 handle limit param to get all systems

r14chandra:


saltgen:
@r14chandra Hey! Could you share the blocker? I just might be able to help

r14chandra:
Currently, Pagination on stage is broken (hence can't verify these changes). As @kgaikwad said the error might be resolved by your PR - #194.

ros/api/v1/hosts.py
```diff
@@ -136,6 +136,9 @@ def get(self):
             .order_by(*sort_expression)
         )
         count = query.count()
+        # a fix to get all the systems
```
kgaikwad, line 139:
```suggestion
        # NOTE: Override limit value to get all the systems when it is -1
```

kgaikwad (changes requested):
Tested! Working as expected.
Added small inline comments.

ros/api/v1/hosts.py
r14chandra: Changes pushed!

kgaikwad (approved):
LGTM :+1:

kgaikwad:
Thanks @r14chandra!

PR #200, opened 2022-03-21
RHIROS-573 add migrations, fix consumers

r14chandra:


kgaikwad (changes requested):
Hello @r14chandra - thank you for the submitting changes.
Added inline comments.
Migrations need to be rebuild and rearrange so that it will be easy to understand and less error prone. I think after rearranging the things in separate migrations we won't require `transaction per migration` .
I tried to build new migrations. If possible could you please separate out consumer side changes into separate commits then I can easy cherry-pick those into my branch.

kgaikwad:
Closing this PR on behalf of https://github.com/RedHatInsights/ros-backend/pull/207
Cherry-picked consumer change commit to that PR. Rest changes will be track into #207 
Thank you @r14chandra!

PR #205, opened 2022-03-25
RHIROS - #532 Implement filtering on OS column

apuntamb:


kgaikwad:
@apuntamb - could you correct issue number to [RHIROS-532](https://issues.redhat.com/browse/RHIROS-532) instead of RHIROS-533?

kgaikwad (changes requested):
@apuntamb - Could you please include simple test?

ros/api/v1/hosts.py
```diff
@@ -182,6 +201,14 @@ def build_system_filters():
             filters.append(System.state.in_(modified_states))
         else:
             filters.append(System.state.in_(SYSTEM_STATES_EXCEPT_EMPTY))
+        if operating_systems := request.args.getlist('os'):
+            modified_operating_systems = []
+            for os in operating_systems:
+                os = os.upper()
+                if os not in OS_VERSIONS.keys():
```
r14chandra, line 208: Here, we're getting `os` as `RHEL X.Y` and `os_versions` keys as 'X.Y'. So, this condition is always true. 

Screenshot from 2022-04-07 13-32-33

r14chandra (changes requested):

ros/api/v1/hosts.py
```diff
@@ -182,6 +201,14 @@ def build_system_filters():
             filters.append(System.state.in_(modified_states))
         else:
             filters.append(System.state.in_(SYSTEM_STATES_EXCEPT_EMPTY))
+        if operating_systems := request.args.getlist('os'):
+            modified_operating_systems = []
+            for os in operating_systems:
+                os = os.upper()
```
r14chandra, line 207: `os` value is already in uppercase, we can remove `upper` function and use `split` instead to get the numerical value 
```suggestion
                os = os.split(" ")[1]
```

ros/api/v1/hosts.py
apuntamb: The motivation here is to pass `/systems?os=7.0`

ros/api/v1/hosts.py
apuntamb: If we are passing `/systems?os=7.0`, then there should be no need of either of these.

ros/api/v1/hosts.py
r14chandra: But on openapi it's `RHEL 7.0`

apuntamb: Updating it in openapi spec here: https://github.com/RedHatInsights/ros-backend/pull/205/files#diff-80c724753dbc2f811ac54a576c78bb6d32e37e6e76fd8e138252a3a151051992R641-R656

apuntamb: can you retest now?

r14chandra: But getting error for every value
Screenshot from 2022-04-07 13-32-33

r14chandra (approved):
Tested,working as expected 💯! Thanks @apuntamb!

ros/openapi/openapi.json
```diff
@@ -624,6 +627,36 @@
                         ]
                     }
                 }
+            },
+            "filterOSParam": {
+                "name": "os",
+                "in": "query",
+                "required": false,
+                "description": "Filter system with the given os version",
```
kgaikwad, line 635:
```suggestion
                "description": "Filter system with the given RHEL version",
```
As you are listing RHEL versions` values and not full version names, will it be good to indicate it in description?

ros/openapi/openapi.json
```diff
@@ -72,6 +72,9 @@
                     },
                     {
                         "$ref": "#/components/parameters/filterStateParam"
+                    },
+                    {
+                        "$ref": "#/components/parameters/filterOSParam"
```
kgaikwad, line 77: Just an observation -
`Advisor` supports filter called `rhel_version` where it accepts version values example - 7.3, 8.5
whereas Patch > Systems supports filter called `os` where it accepts full version name example - RHEL 7.3,RHEL 8.5
Here, you are using filter name as `os` and accepts version values example - 7.3, 8.5

Secondly, on UI side both services are showing full name on filter chip whereas with PR https://github.com/RedHatInsights/ros-frontend/pull/110 change it shows just version value.


For reference attaching screenshots:

Advisor Systems -
Screenshot from 2022-04-08 14-45-56

Patch > Systems -
Screenshot from 2022-04-08 14-47-07

Code functionality wise looks good but I would like to know opinions on which way we would like to keep this behavior?

kgaikwad, line 77: Just an observation -
`Advisor` supports filter called `rhel_version` where it accepts version values example - 7.3, 8.5
whereas Patch > Systems supports filter called `os` where it accepts full version name example - RHEL 7.3,RHEL 8.5
Here, you are using filter name as `os` and accepts version values example - 7.3, 8.5

Secondly, on UI side both services are showing full name on filter chip whereas with PR https://github.com/RedHatInsights/ros-frontend/pull/110 change it shows just version value.


For reference attaching screenshots:

Advisor Systems -
Screenshot from 2022-04-08 14-45-56

Patch > Systems -
Screenshot from 2022-04-08 14-47-07
-45-56.png…]()

kgaikwad (changes requested):

ros/openapi/openapi.json
apuntamb: Since API response has field name as `os` so, I would prefer keeping it as `os` here as well. For the corresponding value, we can keep it as `RHEL x.y` but, my only query is what if user manually tries to hit the API and enters `RHELx.y` without a space? Keeping that in mind, I had kept these values as `x.y` and not `RHEL x.y`. Let me know if that is not a concern, I can update it to `RHEL x.y`.

ros/api/v1/hosts.py
```diff
@@ -21,6 +21,25 @@
...
+    '7.7': {"name": "RHEL", "major": 7, "minor": 7},
+    '7.8': {"name": "RHEL", "major": 7, "minor": 8},
+    '7.9': {"name": "RHEL", "major": 7, "minor": 9},
+    '8.0': {"name": "RHEL", "major": 8, "minor": 0},
+    '8.1': {"name": "RHEL", "major": 8, "minor": 1},
+    '8.2': {"name": "RHEL", "major": 8, "minor": 2},
+    '8.3': {"name": "RHEL", "major": 8, "minor": 3},
+    '8.4': {"name": "RHEL", "major": 8, "minor": 4},
+    '8.5': {"name": "RHEL", "major": 8, "minor": 5}
+}
```
saltgen, line 41: @apuntamb Shall we move this into the `constants.py` file?

ros/openapi/openapi.json
PreetiW: I agree with @kgaikwad it will be good to be consistent with other services and have param name as `os` with full values `RHEL x.y` and showing full name on the frontend side in the filter chip 👍🏻

apuntamb: done.

apuntamb: done.

r14chandra (approved):

kgaikwad (approved):
Tested latest change, working as expected!

kgaikwad:
@saltgen - do you have any additional comment(s)?

ros/api/v1/hosts.py
apuntamb: @saltgen For now, let's keep this constant along with the others in the same file. We can consider refactoring in a separate PR.

kgaikwad:
Thank you @apuntamb!

PR #208, opened 2022-03-28
RHIROS-597 | Executive Report API changes

saltgen:
### Changes:

1. Route and handler class for executive report response
2. Util functions

### Additional:

- These API changes should be independent of db schema updates, that said these changes should be merged
only after #203 

Sample Records:

```
-[ RECORD 1 ]----+--------------------------------------------------
id               | x
account_id       | x
inventory_id     | xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
cloud_provider   | aws
display_name     | xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
fqdn             | xxxxxxxxxxxxxxxxxxxxxxxxxxxxx
instance_type    | t2.micro
state            | Undersized
stale_timestamp  | 2022-05-06 12:07:15.952262+00
region           | us-east-1
operating_system | {"name": "RHEL", "major": 7, "minor": 7}
cpu_state        | {CPU_UNDERSIZED,CPU_UNDERSIZED_BY_PRESSURE}
io_state         | {IO_UNDERSIZED_BY_PRESSURE}
memory_state     | {MEMORY_UNDERSIZED,MEMORY_UNDERSIZED_BY_PRESSURE}

-[ RECORD 2 ]----+--------------------------------------------------
id               | x
account_id       | x
inventory_id     | xxxxxxxxxxxxxxxxxxxxxxxxx
cloud_provider   | aws
display_name     | xxxxxxxxxxxxxxxxxxxxx
fqdn             | xxxxxxxxxxxxxxxxxxxxxxxxxxxxx
instance_type    | xxxxxxxxxxxxxx
state            | Undersized
stale_timestamp  | 2022-05-06 12:08:03.98634+00
region           | us-east-1
operating_system | {"name": "RHEL", "major": 8, "minor": 4}
cpu_state        | {CPU_UNDERSIZED}
io_state         | 
memory_state     | {MEMORY_UNDERSIZED}

-[ RECORD 3 ]----+--------------------------------------------------
id               | x
account_id       | x
inventory_id     | xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
cloud_provider   | aws
display_name     | xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

...skipping 1 line
instance_type    | xxxxxxxxxx
state            | Under pressure
stale_timestamp  | 2022-05-06 12:08:40.739415+00
region           | xxxxxxxxxxxxxxxxxxxxx
operating_system | {"name": "RHEL", "major": 7, "minor": 7}
cpu_state        | {CPU_OVERSIZED,CPU_UNDERSIZED_BY_PRESSURE}
io_state         | {IO_UNDERSIZED_BY_PRESSURE}
memory_state     | {MEMORY_OVERSIZED,MEMORY_UNDERSIZED_BY_PRESSURE}

-[ RECORD 4 ]----+--------------------------------------------------
id               | x
account_id       | x
inventory_id     | xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
cloud_provider   | aws
display_name     | xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
fqdn             | xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
instance_type    | xxxxxxxxxxxxxxxxxxx
state            | Optimized
stale_timestamp  | 2022-05-06 12:10:17.319451+00
region           | xxxxxxxxxxxxxxxxx
operating_system | {"name": "RHEL", "major": 7, "minor": 7}
cpu_state        | 
io_state         | 
memory_state     | 
```

Sample Response:
```

{
    "systems_per_state": {
        "optimized": {
            "count": 1,
            "percentage": 25.0
        },
        "under_pressure": {
            "count": 1,
            "percentage": 25.0
        },
        "undersized": {
            "count": 2,
            "percentage": 50.0
        },
        "oversized": {
            "count": 0,
            "percentage": 0.0
        },
        "waiting_for_data": {
            "count": 0,
            "percentage": 0.0
        },
        "idling": {
            "count": 0,
            "percentage": 0.0
        }
    },
    "conditions": {
        "cpu": {
            "count": 5,
            "percentage": 41.67,
            "undersized": 2,
            "oversized": 1,
            "under_pressure": 2
        },
        "io": {
            "count": 2,
            "percentage": 16.67,
            "undersized": 0,
            "oversized": 0,
            "under_pressure": 2
        },
        "memory": {
            "count": 5,
            "percentage": 41.67,
            "undersized": 2,
            "oversized": 1,
            "under_pressure": 2
        }
    },
    "meta": {
        "total_count": 4,
        "non_optimized_count": 3,
        "conditions_count": 12
    }
}
```

### DIFF
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -1,15 +1,19 @@
-from sqlalchemy import asc, desc, nullslast, nullsfirst
+from sqlalchemy import asc, desc, nullslast, nullsfirst, _or
 from sqlalchemy.types import Float
 from flask import request
 from flask_restful import Resource, abort, fields, marshal_with
 
+from ros.lib.constants import SubStates
 from ros.lib.models import (
     PerformanceProfile, RhAccount, System,
     db, RecommendationRating, PerformanceProfileHistory)
 from ros.lib.utils import (
     is_valid_uuid, identity,
     user_data_from_identity,
-    sort_io_dict, system_ids_by_account)
+    sort_io_dict, system_ids_by_account,
+    count_per_state,
+    calculate_percentage
+)
 from ros.api.common.pagination import (
     build_paginated_system_list_response,
     limit_value,
@@ -376,57 +380,117 @@ class HostHistoryApi(Resource):
 
 class ExecutiveReportAPI(Resource):
 
+    systems_per_state = {
+        "under_pressure": fields.Raw,
+        "undersized": fields.Raw,
+        "oversized": fields.Raw,
+        "waiting_for_data": fields.Raw,
+        "idling": fields.Raw
+    }
+    condtions = {
+        "cpu": fields.Raw,
+        "io": fields.Raw,
+        "memory": fields.Raw
+    }
+    meta = {
+        "total_count": fields.Integer,
+        "non_optimized_count": fields.Integer,
+        "conditions_count": fields.Integer,
+    }
+    report_fields = {
+        "systems_per_state": fields.Nested(systems_per_state),
+        "condtions": fields.Nested(condtions),
+        "meta": fields.Nested(meta)
+    }
+
+    @marshal_with(report_fields)
     def get(self):
+        account_number = identity(request)['identity']['account_number']
+        system_queryset = system_ids_by_account(account_number)
+
+        # System counts
+        all_systems_count = system_queryset.count()
+        under_pressure_systems = count_per_state(system_queryset, {'state': "Under pressure"})
+        undersized_systems = count_per_state(system_queryset, {'state': "Undersized"})
+        oversized_systems = count_per_state(system_queryset, {'state': "Oversized"})
+        awaiting_data_systems = count_per_state(system_queryset, {'state': "Waiting for data"})
+        idling_systems = count_per_state(system_queryset, {'state': "Idling"})
+
+        non_optimized = system_queryset.filter(
+            or_(
+                System.cpu_state is not None,
+                System.io_state is not None,
+                System.memory_state is not None,
+            )
+        )
+
+        # Condition counts
+        non_optimized_count = non_optimized.count()
+        systems_cpu_issues = system_queryset.filter(System.cpu_state is not None).count()
+        cpu_undersized = count_per_state(non_optimized, {'cpu_state': SubStates.UNDERSIZED.value})
+        cpu_under_pressure = count_per_state(non_optimized, {'cpu_state': SubStates.PRESSURE.value})
+        cpu_oversized = count_per_state(non_optimized, {'cpu_state': SubStates.OVERSIZED.value})
+
+        systems_memory_issues = system_queryset.filter(System.memory_state is not None).count()
+        memory_undersized = count_per_state(non_optimized, {'memory_state': SubStates.UNDERSIZED.value})
+        memory_pressure = count_per_state(non_optimized, {'memory_state': SubStates.PRESSURE.value})
+        memory_oversized = count_per_state(non_optimized, {'memory_state': SubStates.OVERSIZED.value})
+
+        systems_io_issues = system_queryset.filter(System.io_state is not None).count()
+        io_undersized = count_per_state(non_optimized, {'io_state': SubStates.UNDERSIZED.value})
+        io_pressure = count_per_state(non_optimized, {'io_state': SubStates.PRESSURE.value})
+        io_oversized = count_per_state(non_optimized, {'io_state': SubStates.OVERSIZED.value})
+
         response = {
             "systems_per_state": {
                 "under_pressure": {
-                    "count": 0,
-                    "percentage": 5.57
+                    "count": under_pressure_systems,
+                    "percentage": calculate_percentage(under_pressure_systems / all_systems_count)
                 },
                 "undersized": {
-                    "count": 1,
-                    "percentage": 7.68
+                    "count": undersized_systems,
+                    "percentage": calculate_percentage(undersized_systems / all_systems_count)
                 },
                 "oversized": {
-                    "count": 7,
-                    "percentage": 8.91
+                    "count": oversized_systems,
+                    "percentage": calculate_percentage(oversized_systems / all_systems_count)
                 },
                 "waiting_for_data": {
-                    "count": 8,
-                    "percentage": 5.57
+                    "count": awaiting_data_systems,
+                    "percentage": calculate_percentage(awaiting_data_systems / all_systems_count)
                 },
                 "idling": {
-                    "count": 8,
-                    "percentage": 5.57
+                    "count": idling_systems,
+                    "percentage": calculate_percentage(idling_systems / all_systems_count)
                 }
             },
             "conditions": {
                 "cpu": {
-                    "count": 23,
-                    "percentage": 20.88,
-                    "undersized": None,
-                    "oversized": None,
-                    "under_pressure": 10
+                    "count": systems_cpu_issues,
+                    "percentage": calculate_percentage(systems_cpu_issues/non_optimized_count),
+                    "undersized": cpu_undersized,
+                    "oversized": cpu_oversized,
+                    "under_pressure": cpu_under_pressure
                 },
                 "io": {
-                    "count": 23,
-                    "percentage": 20.88,
-                    "undersized": 10,
-                    "oversized": 3,
-                    "under_pressure": 10
+                    "count": systems_io_issues,
+                    "percentage": calculate_percentage(systems_io_issues/non_optimized_count),
+                    "undersized": io_undersized,
+                    "oversized": io_oversized,
+                    "under_pressure": io_pressure
                 },
-                "ram": {
-                    "count": 23,
-                    "percentage": 20.88,
-                    "undersized": 10,
-                    "oversized": None,
-                    "under_pressure": 10
+                "memory": {
+                    "count": systems_memory_issues,
+                    "percentage": calculate_percentage(systems_memory_issues/non_optimized_count),
+                    "undersized": memory_undersized,
+                    "oversized": memory_oversized,
+                    "under_pressure": memory_pressure
                 }
             },
             "meta": {
-                "total_count": 55,
-                "non_optimized_count": 23,
-                "conditions_count": 69
+                "total_count": all_systems_count,
+                "non_optimized_count": non_optimized_count,
+                "conditions_count": systems_cpu_issues + systems_memory_issues + systems_io_issues
             }
         }
 
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -381,8 +381,8 @@ class HostHistoryApi(Resource):
 class ExecutiveReportAPI(Resource):
 
     count_and_percentage = {
-          "count": fields.Integer,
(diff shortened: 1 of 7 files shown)


saltgen:
**Note:** Tests via GH Actions will be failing as of now.

ros/api/v1/hosts.py
```diff
@@ -376,57 +380,138 @@ def get(self, host_id):
...
+        non_optimized = system_queryset.filter(
+            or_(
+                System.cpu_state is not None,
+                System.io_state is not None,
+                System.memory_state is not None,
+            ),
+            and_(
+                System.state != 'Optimized',
+                System.state != 'Waiting for data'
+            )
```
saltgen, line 428: @kgaikwad Does this make sense to you? :thinking:

kgaikwad: @saltgen  - instead of these multiple conditions, does it make more sense to check if system state is any of `"Oversized", "Undersized", "Idling", "Under pressure"` then it is non optimized one?

kgaikwad (changes requested):

ros/api/v1/hosts.py
```diff
@@ -1,15 +1,20 @@
-from sqlalchemy import func, asc, desc, nullslast, nullsfirst
+from sqlalchemy import func, asc, desc, or_
```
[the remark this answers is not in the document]
saltgen: @kgaikwad Current version, [here](https://github.com/RedHatInsights/ros-backend/pull/208/files/21ff8f94b5d92a4833632c1557e8c4055ff68679#diff-a8cad03f08edde21a37e001d0e4efbfb39b3fe06591c7d7ade036dbf91705925R1)

ros/lib/utils.py
```diff
@@ -111,8 +112,10 @@ def sort_io_dict(performance_utilization: dict):
     return performance_utilization
 
 
-def system_ids_by_account(account_number):
+def system_ids_by_account(account_number, fetch_records=False):
```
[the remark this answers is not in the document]
saltgen: @kgaikwad IMO we should leverage function overloading as long it does not get too complex, this can keep things simple without adding extra lines of code.

ros/api/v1/hosts.py
saltgen: @kgaikwad Yes that could work. 

My concern is that state and substate data are both coming from an external source, so I think just to be completely sure we should also check the sustates in addition to states.


The resulting query is something like this, it is longer.

```
non_optimized = system_queryset.filter(
            or_(
                System.cpu_state is not None,
                System.io_state is not None,
                System.memory_state is not None,
            ),
            and_(
                System.state == 'Oversized',
                System.state == 'Undersized',
                System.state == 'Idling',
                System.state == 'Under pressure',
            )
        )

```
Thoughts?

kgaikwad: @saltgen  - could you share any scenario where system will hit substates & not main state?

ros/api/v1/hosts.py
```diff
@@ -376,57 +380,142 @@ def get(self, host_id):
...
     def get(self):
+        account_number = identity(request)['identity']['account_number']
+        system_queryset = system_ids_by_account(account_number, fetch_records=True)
+
+        # System counts
+        all_systems_count = system_queryset.count()
+        under_pressure_systems = count_per_state(system_queryset, {'state': "Under pressure"})
+        undersized_systems = count_per_state(system_queryset, {'state': "Undersized"})
+        oversized_systems = count_per_state(system_queryset, {'state': "Oversized"})
+        awaiting_data_systems = count_per_state(system_queryset, {'state': "Waiting for data"})
```
apuntamb, line 420:
```suggestion
        waiting_for_data_systems = count_per_state(system_queryset, {'state': "Waiting for data"})
```

ros/api/v1/hosts.py
apuntamb: Even if states are coming from advisor engine, a system ideally has to hit a main state always. Having said that, IMHO we can think of checking the main state and then consider checking any sub-state.

ros/api/v1/hosts.py
```diff
@@ -376,57 +380,142 @@ def get(self, host_id):
...
+            ),
+            and_(
+                System.state != 'Optimized',
+                System.state != 'Waiting for data'
+            )
+        )
+        non_optimized_count = non_optimized.count()
+
+        # cpu counts
+        total_cpu_issues = sum([
```
apuntamb, line 437: What does the term `issues` signify ?

ros/lib/utils.py
```diff
@@ -132,3 +135,14 @@ def insert_performance_profiles(session, system_id, fields):
         new_entry = model_class(**fields)
         session.add(new_entry)
         session.flush()
+
+
+def count_per_state(queryset, custom_filters: dict):
+    return queryset.filter_by(**custom_filters).count() if queryset else None
```
apuntamb, line 141: In case of `else` condition, do we want to return `None` or something like a `-1` ?

saltgen: @apuntamb I think this data would directly go into the generated pdf, and there would be no use cases for sorting.
@PreetiW Thoughts?

kgaikwad: > In case of `else` condition, do we want to return `None` or something like a `-1` since we are talking about counts here and also for frontend to be able to consume it better ?

Good point @apuntamb!
@saltgen, in this case can't we simply return `0` than `None`?

ros/api/v1/hosts.py
saltgen: Problematic sub-states of cpu component per system

ros/lib/utils.py
PreetiW: @saltgen @kgaikwad yes this data we directly use in the pdf report and for now I have put a check for conditions count data that if it is null then display `0` on the frontend but I agree with @kgaikwad it will be good if backend can send 0 for count and percentage when the data is not available. Let me know your thoughts and accordingly will make changes in the frontend.

tests/test_api_endpoints.py
```diff
@@ -178,6 +180,28 @@ def test_system_rating(
...
+        )
+        with app.app_context():
+            g = db_get_host('ee0b9978-fe1b-4191-8408-cbadbd47f7a3')
+            pprint(g.__dict__)
+        assert response.json['systems_per_state']['idling']['count'] == 1
+        assert response.json['systems_per_state']['idling']['percentage'] == 100
+        assert response.json['conditions']['cpu']['count'] == 2
+        assert response.json['conditions']['io']['count'] == 1
+        assert response.json['conditions']['memory']['count'] == 2
+
```
PreetiW, line 203: @saltgen what you think about including a test case which will have all the data and also wanted to know in case the state or conditions information is not available will backend return the key with some value of the key will not be included in the response?

PreetiW (commented):
Thank you for the PR @saltgen.
I have left few queries as comments.

ros/api/v1/hosts.py
```diff
@@ -376,57 +380,142 @@ def get(self, host_id):
...
                 "under_pressure": {
-                    "count": 0,
-                    "percentage": 5.57
+                    "count": under_pressure_systems,
+                    "percentage": calculate_percentage(under_pressure_systems, all_systems_count)
                 },
                 "undersized": {
-                    "count": 1,
-                    "percentage": 7.68
+                    "count": undersized_systems,
```
PreetiW, line 476: @saltgen once Stefan provides the inputs on the final order in the [jira card](https://issues.redhat.com/browse/RHIROS-516), can you follow the same and send response in the same order for state and conditions please?

tests/test_api_endpoints.py
saltgen: @PreetiW The current test asserts the same as per test system. Also all keys will be returned irrespective of value.

saltgen: @PreetiW The current test asserts the same as per test system. Also all keys will be returned irrespective of value.

ros/api/v1/hosts.py
saltgen: @kgaikwad Based on discussions this is the current query, change pushed.

```
non_optimized = system_queryset.filter(
            and_(
                System.state != 'Optimized',
                System.state != 'Waiting for data'
            )
        )
```

saltgen: @kgaikwad Based on discussions this is the current query, change pushed.

```
non_optimized = system_queryset.filter(
            and_(
                System.state != 'Optimized',
                System.state != 'Waiting for data'
            )
        )
```

ros/api/v1/hosts.py
saltgen: Change pushed

ros/lib/utils.py
saltgen: @kgaikwad @PreetiW Done, `count` and `percentage` is now `0` and `0.0` respectively in such cases.
Change pushed

ros/api/v1/hosts.py
saltgen: @PreetiW I have pushed the change as per discussion, let me know if changes are required yet again.

saltgen:
@kgaikwad @r14chandra @PreetiW  Please find the same response based on an example set of db records, in the description comment, thanks!

ros/api/v1/hosts.py
saltgen: @kgaikwad Latest query,

```
non_optimized = system_queryset.filter(
            or_(
                System.state == 'Oversized',
                System.state == 'Undersized',
                System.state == 'Idling',
                System.state == 'Under pressure'
            )
        )
```

ros/lib/utils.py
saltgen: @kgaikwad @PreetiW Done, `count` and `percentage` is going to be `0` and `0.0` for such cases.
Change pushed

ros/api/v1/hosts.py
```diff
@@ -376,57 +380,145 @@ def get(self, host_id):
...
+        memory_oversized = non_optimized.filter(System.memory_states.any(SubStates.MEMORY_OVERSIZED.value)).count()
+
+        # io counts
+        total_io_issues = sum([
+            len(system.io_states) if system.io_states else 0
+            for system in non_optimized.filter(System.io_states is not None)
+        ])
+        io_undersized = non_optimized.filter(System.io_states.any(SubStates.IO_UNDERSIZED.value)).count()
+        io_pressure = non_optimized.filter(System.io_states.any(SubStates.IO_UNDER_PRESSURE.value)).count()
+        io_oversized = non_optimized.filter(System.io_states.any(SubStates.IO_OVERSIZED.value)).count()
```
kgaikwad, line 464: @saltgen, in my opinion, multiple filter queries and looping on those filtered systems is creating unnecessary complexity here. Can't you simple iterate on existing non_optimized systems ONLY once and calculate summation against each?

kgaikwad (changes requested):

kgaikwad:
Closing this PR on behalf of https://github.com/RedHatInsights/ros-backend/pull/224

kgaikwad:
All unresolved comments will be track in a separate PR.

PR #210, opened 2022-04-04
Bugfix | ROS Prod Deployment 30-03-2022

saltgen:
### Changes:

1. Set CW_ENABLED to str on respective saas-templates
2. Addressed a flake8 warning in routes.py


kgaikwad (approved):
As per the document reference - https://docs.openshift.com/container-platform/4.10/rest_api/template_apis/template-template-openshift-io-v1.html#parameters-2,  LGTM!

PR #213, opened 2022-04-05
Initialization of AWS env vars for logging | Attempt 2

saltgen:
Reference: https://gitlab.cee.redhat.com/insights-platform/saas-templates/-/blob/master/templates/advisor/insights-advisor-api.yaml#L109


apuntamb (approved):
Looks good!

PR #215, opened 2022-04-27
Refs RHIROS-576 - add env params in clowdapp.yaml

kgaikwad:
[Urgent] Need this change to override env vars related to Garbage collector processor for testing and for future use.

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -73,6 +73,10 @@ objects:
         env:
           - name: CLOWDER_ENABLED
             value: ${CLOWDER_ENABLED}
+          - name: GARBAGE_COLLECTION_INTERVAL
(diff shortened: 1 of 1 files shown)


clowdapp.yaml
```diff
@@ -57,6 +57,10 @@ objects:
             value: ${CLOWDER_ENABLED}
           - name: ENABLE_RBAC
             value: "${ENABLE_RBAC}"
+          - name: GARBAGE_COLLECTION_INTERVAL
```
r14chandra, line 60: Should we consider adding these to processor container only?

kgaikwad: Added on processor side only to avoid confusion! Thanks.

saltgen (approved):

kgaikwad:
Merging this PR!

PR #218, opened 2022-04-28
RHIROS-630 | host-inventory as dependency 

digitronik:


clowdapp.yaml
```diff
@@ -11,6 +11,7 @@ objects:
   spec:
     envName: ${ENV_NAME}
     dependencies:
+    - host-inventory
```
r14chandra, line 14: Could you add `ingress` as well

r14chandra (changes requested):

clowdapp.yaml
digitronik: I think host-inventory has ingress as a dependency... In the log, I can see ingress is deployed.

r14chandra (approved):
LGTM 👍🏻

PR #219, opened 2022-05-04
RHIROS-634 add check for systems with no cloud_provider under suggestions

r14chandra:


PreetiW (commented):
@kgaikwad @r14chandra Since we found the root cause of this, I think this PR is no longer needed. Any thoughts?

r14chandra:
I still see the error on prod

r14chandra:
@PreetiW Root cause you mean kafka consumer?

kgaikwad (approved):
To avoid 500 errors on API requests, merging this change as a temporary fix.

PR #220, opened 2022-05-06
RHIROS-689 Added monitoring process for Openshift props

patilsuraj767:
In this PR we are giving each processor a name(processor_name) and adding its reference to PROCESSOR_INSTANCES array. Also we are giving names to each thread where [thread_name](https://github.com/RedHatInsights/ros-backend/pull/220/files#diff-6f48879fb5d9b6c7702c5db83686f993f3fc502e881d30167dd873207967c756L28) is equal to [processor_name](https://github.com/RedHatInsights/ros-backend/pull/220/files#diff-6f48879fb5d9b6c7702c5db83686f993f3fc502e881d30167dd873207967c756R14).
Idea is to simple get the names of all the [running thread](https://github.com/RedHatInsights/ros-backend/pull/220/files#diff-73b796983793ca68d788f97a84b00ef19df5157d5fee7ee854c47ed1133d45f1R160) and check if all PROCESSOR_INSTANCES name is present in list of running thread. if any of the thread is missing/dead we send the 500 response code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ help:
 	@echo "insights-upload-data       upload data to local ingress service for ROS processing"
 
 metadata={"branch_info": {"remote_branch": -1, "remote_leaf": -1}, "bios_uuid": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71", "ip_addresses": ["10.74.255.52", "2620:52:0:4af8:21a:4aff:fe00:a8a"], "fqdn": "vm255-52.gsslab.pnq2.redhat.com", "mac_addresses": ["00:1a:4a:00:0a:8a", "00:00:00:00:00:00"], "satellite_id": -1, "subscription_manager_id": "7846d4fa-6fcc-4b84-aa13-5f12e588ecca", "insights_id": "1d42f242-3828-4a00-8009-c67656c86a51", "machine_id": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71"}
-identity={"identity": {"account_number": "0000001","type": "User","user": {"username": "tuser@redhat.com","email": "tuser@redhat.com","first_name": "test","last_name": "user","is_active": true,"is_org_admin": false, "is_internal": true, "locale": "en_US"},"internal": { "org_id": "000001"}}}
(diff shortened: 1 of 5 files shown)


ros/lib/utils.py
```diff
@@ -146,3 +152,21 @@ def calculate_percentage(numerator, denominator):
         return round((numerator / denominator) * 100, 2)
     else:
         return 0
+
+
+class MonitoringHandler(BaseHTTPRequestHandler):
```
saltgen, line 157: @patilsuraj767 Can we have a sad and happy path unittest for this class with an arbitrary set of threads,
to avoid regression in future and to skip on testing changes locally?

Makefile
```diff
@@ -5,7 +5,7 @@ help:
 	@echo "insights-upload-data       upload data to local ingress service for ROS processing"
 
 metadata={"branch_info": {"remote_branch": -1, "remote_leaf": -1}, "bios_uuid": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71", "ip_addresses": ["10.74.255.52", "2620:52:0:4af8:21a:4aff:fe00:a8a"], "fqdn": "vm255-52.gsslab.pnq2.redhat.com", "mac_addresses": ["00:1a:4a:00:0a:8a", "00:00:00:00:00:00"], "satellite_id": -1, "subscription_manager_id": "7846d4fa-6fcc-4b84-aa13-5f12e588ecca", "insights_id": "1d42f242-3828-4a00-8009-c67656c86a51", "machine_id": "25d6ad97-60fa-4d3e-b2cc-4aa437c28f71"}
-identity={"identity": {"account_number": "0000001","type": "User","user": {"username": "tuser@redhat.com","email": "tuser@redhat.com","first_name": "test","last_name": "user","is_active": true,"is_org_admin": false, "is_internal": true, "locale": "en_US"},"internal": { "org_id": "000001"}}}
+identity={"identity": {"account_number": "0000001", "auth_type": "basic-auth", "type": "User","user": {"username": "tuser@redhat.com","email": "tuser@redhat.com","first_name": "test","last_name": "user","is_active": true,"is_org_admin": false, "is_internal": true, "locale": "en_US"},"internal": { "org_id": "000001"}}}
```
saltgen, line 8: @patilsuraj767 Is this an improvement?

saltgen (commented):
Hey @patilsuraj767 :wave: 
Left a couple of comments, thanks.

Makefile
patilsuraj767: Yes, latest host-inventory image drops the request which does not have `auth_type` set

ros/lib/utils.py
patilsuraj767: Test cases added.

PR #221, opened 2022-05-09
RHIROS-644 - modify log messages from processors

kgaikwad:
- [x]  Unnecessary ROS inventory processor is printing log statements while receiving delete events for systems. In actual very few of them get processed by ROS processor. Modifying log level from info to debug.
- [x]  Renamed to host to system in log messages for consistency.

### DIFF
--- a/ros/processor/inventory_events_consumer.py
+++ b/ros/processor/inventory_events_consumer.py
@@ -101,8 +101,8 @@ class InventoryEventsConsumer:
         host_id = msg['id']
         insights_id = msg['insights_id']
         with app.app_context():
-            LOG.info(
-                '%s - Deleting performance profile records with insights_id %s',
+            LOG.debug(
+                '%s - Received a message for system with insights_id %s',
                 self.prefix,
                 insights_id
             )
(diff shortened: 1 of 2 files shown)


ros/processor/inventory_events_consumer.py
```diff
@@ -101,8 +101,8 @@ def host_delete_event(self, msg):
         host_id = msg['id']
         insights_id = msg['insights_id']
         with app.app_context():
-            LOG.info(
-                '%s - Deleting performance profile records with insights_id %s',
+            LOG.debug(
```
r14chandra, line 104: Don't know why this log was added, but do we really need it.  Also, same for `if rows_deleted > 0:` [here](https://github.com/RedHatInsights/ros-backend/blob/main/ros/processor/inventory_events_consumer.py#L111). Should we replace `>0` with `==1` as only one unique host would be deleted for every delete event ,I suppose. Any thoughts?

kgaikwad: Made condition using `==1`.

kgaikwad: Info level is not required for this log statement while debugging I think it will be helpful.

r14chandra (approved):
Looks good 👍🏻

PR #223, opened 2022-05-11
RHIROS-649 storage for org_id

r14chandra:


migrations/versions/d13d58080d80_.py
```diff
@@ -0,0 +1,30 @@
...
+revision = 'd13d58080d80'
+down_revision = '77fdc857c94f'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('rh_accounts', sa.Column('org_id', sa.Text()))
+    op.add_column('systems', sa.Column('rh_org_id', sa.Text()))
```
kgaikwad, line 22: Why do we need this `rh_org_id` column in `systems` table?

ros/lib/models.py
```diff
@@ -117,6 +118,7 @@ class RhAccount(db.Model):
     __tablename__ = 'rh_accounts'
     id = db.Column(db.Integer, primary_key=True)
     account = db.Column(db.Text, nullable=False)
```
patilsuraj767, line 120: Please relax the `nullable=False` constraint from the `account` as further going we can have entries where account_id is not present.

patilsuraj767: Something more good would be `check (account_id is not null or org_id is not null)`

migrations/versions/d13d58080d80_.py
patilsuraj767: +1

patilsuraj767 (changes requested):

migrations/versions/d13d58080d80_.py
patilsuraj767: +1

r14chandra: For updating db queries we need to use Org Id instead of `account_id` from `systems` table (for eg [host.py](https://github.com/RedHatInsights/ros-backend/blob/main/ros/api/v1/hosts.py#L145)). Both the columns (`rh_org_id ` & `account_id`) would have same values eventually. And later in Level 3, we can delete (optional) `account_id`.

ros/lib/models.py
r14chandra: > Please relax the nullable=False constraint from the account as further going we can have entries where account_id is not present.

This is for Level-3 (handling anemic-tenants). As for now, account **must** be there.

migrations/versions/d13d58080d80_.py
kgaikwad: @r14chandra  - `account_id` from systems table is a foreign key not the actual account number. Actual account number used in method [system_ids_by_account](https://github.com/RedHatInsights/ros-backend/blob/main/ros/api/v1/hosts.py#L139). I don't think required any additional column to `systems` table.


In my opinion either you can rename this foreign key now or later in level 3 to `tenant_id`. 
@r14chandra @patilsuraj767  - any thoughts?

patilsuraj767: Yes `account_id` is foreign key so no need to add `rh_org_id` is systems table. 
Also, I agree we should rename it to `tenant_id` but we will also need rename the method to `system_ids_by_tenant_id`. I think all of the refactoring and renaming can be done in level 3

r14chandra: > `account_id` from systems table is a foreign key not the actual account number. Actual account number used in method [system_ids_by_account](https://github.com/RedHatInsights/ros-backend/blob/main/ros/api/v1/hosts.py#L139). I don't think required any additional column to `systems` table.

Yes, I'm aware of it. That's why I said `rh_org_id & account_id` would have same values eventually. 
As you both said we don't need to add the column then on what basis we would be querying db. I might be missing out something.

kgaikwad: For now, it would be good to keep just `org_id` info in `rh_accounts` table and later simply rename foreign key i.e `account_id` column in `systems` table.

patilsuraj767 (approved):
LGTM :+1:

migrations/versions/d13d58080d80_.py
```diff
@@ -0,0 +1,29 @@
...
+revision = 'd13d58080d80'
+down_revision = 'e65ffbc90def'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('rh_accounts', sa.Column('org_id', sa.Text()))
+    op.create_unique_constraint('uq_org_id', 'rh_accounts', ['org_id'])
```
kgaikwad, line 22:
```suggestion
    op.create_unique_constraint('uq_rh_accounts_org_id', 'rh_accounts', ['org_id'])
```

Refer - https://docs.sqlalchemy.org/en/14/core/constraints.html#configuring-a-naming-convention-for-a-metadata-collection

kgaikwad (changes requested):

kgaikwad (approved):
Awesome! Thank you @r14chandra :+1:

PR #224, opened 2022-05-12
Refs RHIROS-597 | executive api change

kgaikwad:


kgaikwad:
Both jobs are green so merging this change.

PR #225, opened 2022-05-17
RHIROS-662 - fix total system count in executive report

kgaikwad:


ros/api/v1/hosts.py
```diff
@@ -412,9 +412,18 @@ class ExecutiveReportAPI(Resource):
...
         system_queryset = system_ids_by_account(account_number, fetch_records=True)
+        systems_with_performance_record_query = (
+            db.session.query(PerformanceProfile, System, RhAccount)
+            .join(System, System.id == PerformanceProfile.system_id)
+            .join(RhAccount, RhAccount.id == System.account_id)
+            .filter(
+                PerformanceProfile.system_id.in_(
+                    system_ids_by_account(account_number).subquery())
+            )
+        )
```
r14chandra, line 423: Can we query something like this?

```suggestion
        systems_with_performance_record_query = (
            (db.session.query(PerformanceProfile.system_id)
            .filter(PerformanceProfile.system_id.in_(system_ids_by_account(account_number).subquery())))
            )
        )
```

kgaikwad: Good catch! :+1: 
Yes, it will be better than loading all columns from these three tables as don't require any additional info.

PR #226, opened 2022-05-18
RHIROS-651 populate org_id using platform_metadata

r14chandra:


ros/processor/inventory_events_consumer.py
```diff
@@ -138,7 +138,8 @@ def process_system_details(self, msg):
             try:
                 account = get_or_create(
                     db.session, RhAccount, 'account',
-                    account=host['account']
+                    account=host['account'],
+                    org_id=msg['platform_metadata']['org_id']
```
kgaikwad, line 142:
```suggestion
                    org_id=msg['platform_metadata'].get('org_id')
```

kgaikwad (changes requested):
Thank you @r14chandra! Added one small inline comment. 
Do you also need to handle the same in engine processor at [line L112](https://github.com/RedHatInsights/ros-backend/blob/main/ros/processor/insights_engine_result_consumer.py#L112)?

Rest looks good!

r14chandra:
> Do you also need to handle the same in engine processor at [line L112](https://github.com/RedHatInsights/ros-backend/blob/main/ros/processor/insights_engine_result_consumer.py#L112)?
> 

Didn't get that. Did you mean adding `org_id` there?

kgaikwad:
> Didn't get that. Did you mean adding `org_id` there?

Yes! As ROS is creating rh_account record inside engine processor too.

r14chandra:
> Yes! As ROS is creating rh_account record inside engine processor too.

I don't know under which scenario it would create `rh_accounts` object (under engine processor) but while testing locally it worked fine. Also, we're not getting `org_id` value from engine.

patilsuraj767:
@r14chandra 
**Scenario** - Assume for some reason ROS is facing delay in processing messages from inventory-events topic because of which advisor has first run the rule on report and pushed the message on ENGINE_RESULT_TOPIC where our `insights_engine_result_consumer` processor is listening to. In this case ROS will first read message from ENGINE_RESULT_TOPIC instead of INVENTORY_EVENTS_TOPIC. To handle this situation we are trying to create rh_accounts object (if it already does not exist in DB) under engine processor.
In level 2 engine will also need to provide `org_id`.

r14chandra:
> @r14chandra **Scenario** - Assume for some reason ROS is facing delay in processing messages from inventory-events topic because of which advisor has first run the rule on report and pushed the message on ENGINE_RESULT_TOPIC where our `insights_engine_result_consumer` processor is listening to. In this case ROS will first read message from ENGINE_RESULT_TOPIC instead of INVENTORY_EVENTS_TOPIC. To handle this situation we are trying to create rh_accounts object (if it already does not exist in DB) under engine processor. In level 2 engine will also need to provide `org_id`.

Great! Thanks @patilsuraj767 for explaining. So to tackle this, Is it fine to get `org_id` by using [identity](https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py#L52) method as we don't get it from engine currently.

patilsuraj767:
@r14chandra We cannot use [identity](https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py#L52) as this method is meant for API call which has a request object. 
If currently engine is not providing org_id then eventually they will need to provide it at some point in future. Meanwhile what we can do is check **if** we are getting org_id in ENGINE_RESULT_TOPIC **else** use translation API to get the org_id using account_number. 

Translation API - https://backoffice-proxy.apps.ext.spoke.preprod.us-west-2.aws.paas.redhat.com/docs/#/default/accountMapFromEbsID
More details related to translation API is available in internal org_id migration doc.

r14chandra:
> check if we are getting org_id in ENGINE_RESULT_TOPIC
No we're not getting.

patilsuraj767:
Yeah, but whenever we will start getting we will directly have it from topic, as workaround currently it will execute **else block** that is use translation API.

saltgen (commented):
@r14chandra Do we need respective changes in the Makefile?

r14chandra:
> @r14chandra Do we need respective changes in the Makefile?

No we don't.

kgaikwad:
Thanks @r14chandra!
Thank you @patilsuraj767 and @saltgen for review!

PR #227, opened 2022-05-23
RHIROS-652 add org-id-populator job

r14chandra:
Merge this after #226 
NOTE: Needs DEV-PROD team to run this job on stage and prod.

PR #228, opened 2022-05-23
RHIROS-653 rename account_id to tenant_id, update db queries

r14chandra:
Merge after this PR - #227


ros/api/v1/call_to_action.py
```diff
@@ -1,15 +1,15 @@
...
+    identity, system_ids_by_org_id)
 from ros.lib.models import (
     PerformanceProfile, db)
 
 
 class CallToActionApi(Resource):
     def get(self):
         account_number = identity(request)['identity']['account_number']
-        system_query = system_ids_by_account(account_number).filter(PerformanceProfile.number_of_recommendations > 0)
+        system_query = system_ids_by_org_id(account_number).filter(PerformanceProfile.number_of_recommendations > 0)
```
kgaikwad, line 12:
```suggestion
        system_query = system_ids_by_org_id(org_id).filter(PerformanceProfile.number_of_recommendations > 0)
```

Method  `system_ids_by_org_id` accepts `org_id` whereas you passed `account_number`. 
Similarly here also you need to fetch `org_id` value from identity header?

ros/api/v1/hosts.py
```diff
@@ -57,12 +57,12 @@
 
 class IsROSConfiguredApi(Resource):
     def get(self):
-        account_number = identity(request)['identity']['account_number']
+        org_id = identity(request)['identity']['internal']['org_id']
```
kgaikwad, line 60:
```suggestion
        org_id = org_id_from_identity_header(request)              
```

Almost everywhere ros-backend is parsing and fetch `org_id` value using nested keys fron identity header. 
Will it be good idea to encapsulate ` identity(request)['identity']['internal']['org_id']` logic into a new method [here](https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py).
After promoting `org_id` field as top level field in identity header, one needs to update code ONLY at one place.

kgaikwad (changes requested):
Thank you @r14chandra. Added two inline comments!

ros/api/v1/call_to_action.py
r14chandra: Oh I missed this 🤦🏻‍♀️.

ros/api/v1/hosts.py
r14chandra: Yeah good idea 👍🏻

kgaikwad: @r14chandra  - Please go through slack conversation - https://ansible.slack.com/archives/C03B7U5AE67/p1655120034820549?thread_ts=1655111676.332289&cid=C03B7U5AE67
This change i.e org_is as top level field is already available in stage. 
Could you also use top level key to get org_id change?

ros/api/v1/recommendations.py
```diff
@@ -51,9 +51,9 @@ def get(self, host_id):
 
         filter_description = request.args.get('description')
 
-        account_query = db.session.query(RhAccount.id).filter(RhAccount.account == ident['account_number']).subquery()
+        tenant_query = db.session.query(RhAccount.id).filter(RhAccount.org_id == ident['internal']['org_id']).subquery()
```
kgaikwad, line 54:
```suggestion
        tenant_query = db.session.query(RhAccount.id).filter(RhAccount.org_id == ident['org_id']).subquery()
```

kgaikwad (approved):
Thanks @r14chandra  for resolving all comments.
Changes look good to me.

PR #233, opened 2022-06-17
RHIROS-658 not null constraint for rh_accounts's org_id

r14chandra:


migrations/versions/77a3ca142468_.py
```diff
@@ -0,0 +1,42 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('rh_accounts', 'account',
+               existing_type=sa.TEXT(),
+               nullable=True)
+    op.drop_constraint('rh_accounts_account_key', 'rh_accounts', type_='unique')
+    op.drop_constraint('uq_rh_accounts_org_id', 'rh_accounts', type_='unique')
+    op.create_unique_constraint('uq_rh_accounts', 'rh_accounts', ['account', 'org_id'])
+    op.drop_constraint('rh_accounts_account_check', 'rh_accounts', type_='check')
+    op.create_check_constraint('ck_rh_accounts', 'rh_accounts', 'NOT(account IS NULL OR org_id IS NULL)')
```
kgaikwad, line 28: As you have set `account` column as nullable, do we still need to consider `account` column inside condition?

kgaikwad (changes requested):

migrations/versions/77a3ca142468_.py
r14chandra: oh that code was auto-generated. I had already removed 🤔.

migrations/versions/77a3ca142468_.py
```diff
@@ -0,0 +1,42 @@
...
+down_revision = 'a983b8b1865d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('rh_accounts', 'account',
+               existing_type=sa.TEXT(),
+               nullable=True)
```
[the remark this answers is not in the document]
kgaikwad: @r14chandra  - 
Does that mean we don't need update account `column` as nullable?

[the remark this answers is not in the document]
r14chandra: I've set it as not null already. Later in Level 3 we need to set it as nullable=true.

migrations/versions/77a3ca142468_.py
kgaikwad: @r14chandra - so what is expected condition here?

r14chandra: account and org_id should not be null.

r14chandra: > As you have set `account` column as nullable, do we still need to consider `account` column inside condition?

It is set as not null as earlier. I just removed `nullable=false` and added the check contraint for both the columns.

kgaikwad (approved):
LGTM!

PR #240, opened 2022-07-06
RHIROS-736 | System record w/o cloud_provider and/or instance_type

saltgen:
- Added a condition to skip looking for suggestion for systems w/o cloud_provider, instance_type values
- Added an unitest
- Cleaned up the test module a bit

### DIFF
--- a/ros/api/v1/recommendations.py
+++ b/ros/api/v1/recommendations.py
@@ -59,6 +59,17 @@ class RecommendationsApi(Resource):
             abort(404, message="host with id {} doesn't exist"
                   .format(host_id))
 
+        # cloud_provider & instance_type are necessary for suggestions
+        if (
+                system.cloud_provider is None
+                or system.instance_type is None
+        ):
+            return {
+                'inventory_id': system.inventory_id,
+                'data': [],
+                'meta': {'count': 0}
+            }
+
         profile = PerformanceProfile.query.filter_by(system_id=system.id).first()
         if not profile:
             abort(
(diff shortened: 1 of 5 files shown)


ros/api/v1/recommendations.py
```diff
@@ -59,6 +59,17 @@ def get(self, host_id):
             abort(404, message="host with id {} doesn't exist"
                   .format(host_id))
 
+        # cloud_provider & instance_type are necessary for suggestions
+        if (
+                system.cloud_provider is None
+                or system.instance_type is None
+        ):
+            return {
+                'inventory_id': system.inventory_id,
```
r14chandra, line 68: Will this work with frontend?

saltgen: It should, we do have scenarios where-in there are no suggestions w.r.t. a system.
@PreetiW Could you confirm?

r14chandra: Yes there is. For optimized state.

r14chandra: I don't know if this is correct to show others states having no cloud_provider or instance_type as `no suggestion`.

saltgen: @kgaikwad Could you re-confirm on this?

kgaikwad: I am not in favor of sending same result like optimized state when there is missing data. @saltgen  could you explain how frontend can make difference between these two situations?

saltgen: @kgaikwad Atm there should be no difference, if `cloud_provider` info is missing we are treating the system as non-ROS.
Any suggestions if there should be any changes on BE and/or FE?

saltgen: @kgaikwad In case we don't need this, shall I close this out?

r14chandra: +1 for closing this. I don't have any other opinion.

saltgen: Closing this out along with the JIRA card

PR #244, opened 2022-07-13
RHIROS-744 use kafka config values from clowder

r14chandra:


ros/lib/config.py
```diff
@@ -34,7 +48,8 @@ def get_logger(name):
     DB_HOST = LoadedConfig.database.hostname
     DB_PORT = LoadedConfig.database.port
     METRICS_PORT = LoadedConfig.metricsPort
-    INSIGHTS_KAFKA_ADDRESS = LoadedConfig.kafka.brokers[0].hostname + ":" + str(LoadedConfig.kafka.brokers[0].port)
+    KAFKA_BROKER = LoadedConfig.kafka.brokers[0]
+    INSIGHTS_KAFKA_ADDRESS = KAFKA_BROKER.hostname + ":" + str(KAFKA_BROKER.port)
```
kgaikwad, line 52: In my opinion, Can we print this INSIGHTS_KAFKA_ADDRESS?

kgaikwad (approved):

kgaikwad:
Added one small inline comment. I think this can be handled separately. 
For now, merging this change to check stage result as managed kafka tests kicked off against stage.

PR #248, opened 2022-07-14
RHIROS-705 remove account from api response and openapi

r14chandra:


kgaikwad (approved):
LGTM :+1:

kgaikwad:
@r14chandra  - As part of this change, could you please update frontend tests as well? 
See this test file - https://github.com/RedHatInsights/ros-frontend/blob/main/src/Components/Reports/UtilTestData.js

PR #250, opened 2022-07-25
Fixes #RHIROS-759 - refactor executive_report API for SQL queries

kgaikwad:
Note that from local setup I can see current logic uses 14 SQL count queries and 3 SQL select queries.
With this commit, now it uses 7 count queries and 1 SQL select query.

Any suggestions on how we can make this code more efficient as currently I am iterating on all non optimized systems?

### DIFF
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -1,4 +1,4 @@
-from sqlalchemy import asc, desc, nullslast, nullsfirst, or_
+from sqlalchemy import asc, desc, nullslast, nullsfirst
 from sqlalchemy.types import Float
 from flask import request
 from flask_restful import Resource, abort, fields, marshal_with
@@ -420,7 +420,7 @@ class ExecutiveReportAPI(Resource):
         )
 
         # System counts
-        all_systems_count = systems_with_performance_record_query.count()
+        total_systems = systems_with_performance_record_query.count()
         optimized_systems = count_per_state(system_queryset, {'state': "Optimized"})
         under_pressure_systems = count_per_state(system_queryset, {'state': "Under pressure"})
         undersized_systems = count_per_state(system_queryset, {'state': "Undersized"})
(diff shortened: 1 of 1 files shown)


ros/api/v1/hosts.py
```diff
@@ -429,108 +429,103 @@ def get(self):
...
-        io_oversized = -1
-
-        all_conditions_count = total_cpu_issues + total_memory_issues + total_io_issues
+        non_optimized_count = 0
+        total_keys = ['cpu', 'memory', 'io']
+        state_keys = ['undersized', 'under_pressure', 'oversized']
+        totals = dict.fromkeys(total_keys, 0)
+        cpu_states_dict = dict.fromkeys(state_keys, 0)
+        memory_states_dict = dict.fromkeys(state_keys, 0)
+        io_states_dict = dict.fromkeys(state_keys, 0)
```
saltgen, line 441:
```suggestion
        state_keys_init = dict.fromkeys(state_keys, 0)
        cpu_states_dict = memory_states_dict = io_states_dict = state_keys_init
```

saltgen (commented):
After profiling the code I can see a stark difference in performance and function calls! :confetti_ball: 

### Existing code:

`91283 function calls (88240 primitive calls) in 0.097 seconds`

### Post changes:
`30591 function calls (29771 primitive calls) in 0.037 seconds`

These were done with 2 idle and an optimized system in DB.

Thanks for the changes @kgaikwad!

Just added one minor suggestion.

kgaikwad:
Awesome! Thank you @saltgen for sharing results.
+ adding @ShivamPandya on PR as I requested him to share observations from ephemeral env with larget set of systems.(example. >=50 ROS systems)

ros/api/v1/hosts.py
kgaikwad: @saltgen - I intentionally avoided assigning same value to these variables as here I am assigning `dict` objects which come under mutable objects.

ShivamPandya:
@kgaikwad  Thanks for the reference. This were my observations:

Regular observations are stored in _avgLatest.txt_ file and observations after applying changes from this PR are stored in _avgPR250.txt_ file.

I have included around 5 iterations of curl request and have stored time taken by each request. Reason for this is to get an average value as internet speed at any moment can influence the results.

kgaikwad:
@ShivamPandya @saltgen - Merging it.

PR #251, opened 2022-07-25
RHIROS-687 | Executive Report v2 changes

saltgen:
### Changes:
1. Added util methods for creating the instance type highlights dict, generating highlight descriptions
2. Added method for generating description per instance type
3. Updated `constants.py` with AWS instance types
4. `openapi.json` updated

### Additional:
- A Prometheus logger has been added in generate_highlight_description();
  logs a count whenever there is a lookup error in `aws_instance_types.py`


saltgen:
@kgaikwad Pushed changes for considering systems w/o profiles only in the `current` section

saltgen:
@kgaikwad  Since this aws instance types file is static, I found it best to copy it's contents into another file. I agree constants probably is not the appropriate option.

I'll create a file with the name `aws_instance_types.py` and copy the dict content there. 

Regarding the sync mechanism I'm not too sure if we can update this same file, but lets discuss soon.

ros/api/v1/hosts.py
```diff
@@ -472,6 +482,27 @@ def get(self):
...
+            for system in system_queryset
+            for record in performance_profiles_by_system_id(system.id, PerformanceProfile)
+            if record.report_date.date() == upload_date
+        ]
+
+        historical_performance_profiles = [
+            record
+            for system in system_queryset
+            for record in performance_profiles_by_system_id(system.id, PerformanceProfileHistory)
+        ]  # Systems older than 7 days/stale systems are considered here
```
PreetiW, line 501: @saltgen do you know if we are going to include today's data as well while considering this? or 45 days of data till yesterday as in above section we are considering yesterday's data?

ros/lib/utils.py
```diff
@@ -185,3 +193,62 @@ def log_request(self, code='-', size='-'):
...
+    cloud = 'AWS'  # Update when multi cloud provider support is added
+
+    description_text = f'{cpu_type} instance with {num_vcpus} vCPUs ' \
+                       f'and {ram_gb} of RAM, running on {cloud} ' \
+                       f'{cloud_regions} regions'
+
+    return description_text
+
+
+def highlights_instance_types(queryset, highlight_type='historical'):
```
PreetiW, line 218: @saltgen I feel it will be good if we can remove the `historical` as the default value for this function and make it accept 3 values to make it more readable, if you have any other suggestions for same do let me know 🙌🏻

PreetiW (commented):
Thank you @saltgen for the PR! did the first round of the review and left few comments on it. Let me know what you think!

ros/lib/utils.py
saltgen: @PreetiW As discussed, we can do some minor refactoring on this one, I'll get to it once I'm back at work, thanks!

ros/api/v1/hosts.py
PreetiW: Based on discussion in the review meeting it was decided that this require no code change. hence resolving this.

PreetiW (approved):
I tested the changes locally and in EE, it is working fine!

PreetiW (approved):
Thanks @saltgen for the PR, except the minor code refactoring for util function everything looks good 💯 

I have tested the changes in EE and locally, everything is working correctly 👍🏻 

image

saltgen:
@PreetiW Thanks for verifying the changes, I have refactored the util function and squashed some redundant commits.

patilsuraj767 (approved):
lgtm, Just observed one behavior not sure if it is right or wrong. If any new user register the system for first time in ros and then hit executive report API then only `instance_types_highlights` for `historical` is populated, `current` and `suggested` returns empty array. It is caused because of condition - [here](https://github.com/RedHatInsights/ros-backend/pull/251/files#diff-a8cad03f08edde21a37e001d0e4efbfb39b3fe06591c7d7ade036dbf91705925R494)

ros/api/v1/hosts.py
```diff
@@ -472,6 +482,27 @@ def get(self):
         io_states_dict['undersized'] = -1
         total_conditions = totals['cpu'] + totals['memory'] + totals['io']
 
+        current_utc_datetime = datetime.utcnow().replace(tzinfo=pytz.utc)
```
kgaikwad, line 485:
```suggestion
       current_utc_datetime = datetime.now(timezone.utc)
```
Reference - https://docs.python.org/3/library/datetime.html#datetime.datetime.now

kgaikwad (changes requested):

saltgen:
> lgtm, Just observed one behavior not sure if it is right or wrong. If any new user register the system for first time in ros and then hit executive report API then only `instance_types_highlights` for `historical` is populated, `current` and `suggested` returns empty array. It is caused because of condition - [here](https://github.com/RedHatInsights/ros-backend/pull/251/files#diff-a8cad03f08edde21a37e001d0e4efbfb39b3fe06591c7d7ade036dbf91705925R494)

Thanks for looking into this @patilsuraj767. Yes, this is an existing edge case for a new user.

Atm we are going ahead with this by design i.e. historical section will be showing all 45 days of data.

We will revisit this later, if necessary.

ros/api/v1/hosts.py
saltgen: The timezone metadata seems to be a bit different for each of the functions, I'm investigating.
I'll push shortly, thanks!
```
datetime.now(timezone.utc)
datetime.datetime(2022, 10, 14, 19, 12, 15, 782847, tzinfo=datetime.timezone.utc)
datetime.utcnow().replace(tzinfo=pytz.utc)
datetime.datetime(2022, 10, 14, 19, 12, 19, 228398, tzinfo=<UTC>)
```

ros/api/v1/hosts.py
```diff
@@ -477,6 +487,32 @@ def get(self):
 
         all_conditions_count = total_cpu_issues + total_memory_issues + total_io_issues
 
+        current_utc_datetime = datetime.utcnow().replace(tzinfo=pytz.utc)
+
+        # target upload date is yesterday's date
```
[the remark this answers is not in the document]
kgaikwad: @saltgen  - This variable is used only once in query while applying filter on `report_date`.
I am fine to keep it as `upload_date`.  I suggested this change because from assignment (`(current_utc_datetime - timedelta(days=1)).date()`) it is confusing why it is being called as `upload_date`.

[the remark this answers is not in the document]
saltgen: @kgaikwad Kept it as `upload_date` as  that is currently the on wards date we are considering for profile uploads in `current` and `suggested` section, would you like to suggest further changes?

ros/api/v1/hosts.py
saltgen: @kgaikwad Change pushed

saltgen:
@kgaikwad @patilsuraj767 Do we need `system_queryset` [here](https://github.com/RedHatInsights/ros-backend/blob/d934f8a8e6925aa257d50812e69c07679263c69f/ros/api/v1/hosts.py#L435-L445)?

I think we can do this with `systems_with_performance_record_queryset` itself since we have state in performance profile model

kgaikwad:
> @kgaikwad @patilsuraj767 Do we need `system_queryset` [here](https://github.com/RedHatInsights/ros-backend/blob/d934f8a8e6925aa257d50812e69c07679263c69f/ros/api/v1/hosts.py#L435-L445)?
> 
> I think we can do this with `systems_with_performance_record_queryset` itself since we have state in performance profile model

@saltgen  +1, make sense to me!

ros/lib/utils.py
```diff
@@ -185,3 +193,61 @@ def log_request(self, code='-', size='-'):
...
+    num_vcpus = INSTANCE_TYPES[instance_type]['extra']['vcpu']
+    ram_gb = INSTANCE_TYPES[instance_type]['extra']['memory']
+    cloud_regions = INSTANCE_TYPES[instance_type]['extra']['regionCode']
+    cloud = 'AWS'  # Update when multi cloud provider support is added
+
+    description_text = f'{cpu_type} instance with {num_vcpus} vCPUs ' \
+                       f'and {ram_gb} of RAM, running on {cloud} ' \
+                       f'{cloud_regions} regions'
+
+    return description_text
```
kgaikwad, line 214:
```suggestion
   system_instance_type = INSTANCE_TYPES.get(instance_type)
    if system_instance_type is None:
        # logging lookup failure on Prometheus
        ec2_instance_lookup_failures.labels(reporter='API Events').inc()
        return 'NA'
    instance_type_extra = system_instance_type.get('extra')
    
    return ("{cpu_type} instance with {num_vcpus} vCPUs " \
                "and {ram_gb} of RAM, running on {cloud}" \
                "{cloud_regions} regions").format(cpu_type=instance_type_extra.get('physicalProcessor'),
                num_vcpus=instance_type_extra.get('vcpu'), ram_gb=instance_type_extra.get('memory'),
                cloud_regions=instance_type_extra.get('regionCode'), cloud= 'AWS' ) # Update when multi cloud provider support is added

```

Instead of reading object multiple times from `INSTANCE_TYPES`, will it be more efficient to capture respective instance_type object in single variable?

ros/lib/utils.py
```diff
@@ -185,3 +193,61 @@ def log_request(self, code='-', size='-'):
...
+        for performance_profile in queryset:
+            try:
+                instance_candidates.append(performance_profile.rule_hit_details[0]['details']['instance_type'])
+            except (IndexError, KeyError):
+                # Systems w/o profiles are being considered in 'current'
+                # Ex: Optimized, Waiting for Data, etc.
+                system_record = db.session.query(System).\
+                    filter_by(id=performance_profile.system_id).first()
+                if system_record:
+                    instance_candidates.append(system_record.instance_type)
```
kgaikwad, line 229: @saltgen  - instead of special handling for keyerror/indexerror, will it be good idea to join performance_profile and system table? 

IIUC, mainly it requires inventory_id, system.state, system.instance_type, system.region, rule_hit_details.
Will it make more sense to load required fields than all fields details?

kgaikwad (changes requested):

ros/lib/utils.py
saltgen: @kgaikwad I ran some profiling with the suggested change,

Suggested change:

```
Timer unit: 1e-06 s
Total time: 8e-06 s
File: /home/nik/workspace/forked/ros-backend/ros/lib/utils.py
Function: generate_highlight_description at line 202
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   202                                           def generate_highlight_description(instance_type):
   203         1          1.0      1.0     12.5      system_instance_type = INSTANCE_TYPES.get(instance_type)
   204         1          1.0      1.0     12.5      if system_instance_type is None:
   205                                                   # logging lookup failure on Prometheus
   206                                                   ec2_instance_lookup_failures.labels(reporter='API Events').inc()
   207                                                   return 'NA'
   208         1          1.0      1.0     12.5      instance_type_extra = system_instance_type.get('extra')
   209                                           
   210         2          3.0      1.5     37.5      description_text = "{cpu_type} instance with {num_vcpus} vCPUs and {ram_gb} of RAM, running on {cloud}{cloud_regions} regions".format(
   211         1          0.0      0.0      0.0          cpu_type=instance_type_extra.get('physicalProcessor'),
   212         1          1.0      1.0     12.5          num_vcpus=instance_type_extra.get('vcpu'),
   213         1          0.0      0.0      0.0          ram_gb=instance_type_extra.get('memory'),
   214         1          0.0      0.0      0.0          cloud_regions=instance_type_extra.get('regionCode'),
   215         1          1.0      1.0     12.5          cloud='AWS')  # Update when multi cloud provider support is added
   216                                           
   217         1          0.0      0.0      0.0      return description_text
```

Existing change:

```
Timer unit: 1e-06 s
Total time: 8e-06 s
File: /home/nik/workspace/forked/ros-backend/ros/lib/utils.py
Function: generate_highlight_descriptione at line 220
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   220                                           def generate_highlight_descriptione(instance_type):
   221         1          5.0      5.0     62.5      if instance_type not in INSTANCE_TYPES.keys():
   222                                                   # logging lookup failure on Prometheus
   223                                                   ec2_instance_lookup_failures.labels(reporter='API Events').inc()
   224                                                   return 'NA'
   225                                           
   226         1          1.0      1.0     12.5      cpu_type = INSTANCE_TYPES[instance_type]['extra']['physicalProcessor']
   227         1          0.0      0.0      0.0      num_vcpus = INSTANCE_TYPES[instance_type]['extra']['vcpu']
   228         1          0.0      0.0      0.0      ram_gb = INSTANCE_TYPES[instance_type]['extra']['memory']
   229         1          0.0      0.0      0.0      cloud_regions = INSTANCE_TYPES[instance_type]['extra']['regionCode']
   230         1          0.0      0.0      0.0      cloud = 'AWS'  # Update when multi cloud provider support is added
   231                                           
   232         1          1.0      1.0     12.5      description_text = f'{cpu_type} instance with {num_vcpus} vCPUs ' \
   233                                                                  f'and {ram_gb} of RAM, running on {cloud} ' \
   234                                                                  f'{cloud_regions} regions'
   235                                           
   236         1          1.0      1.0     12.5      return description_text
```
As we can  see the total time remains roundabout the same, here are the key points

1.  `system_instance_type = INSTANCE_TYPES.get(instance_type)` is a good suggestion, the `in` lookup is expensive
2.  Fetching the `instance type['extra']` multiple times is not a problem as Python loads it only once in memory
3.  IMHO we should push towards f-strings going forward instead of format, as it is faster and easier to read

Here is the  profiling including the suggestion,
```
Timer unit: 1e-06 s
Total time: 6e-06 s
File: /home/nik/workspace/forked/ros-backend/ros/lib/utils.py
Function: generate_highlight_descriptione at line 220
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   220                                           def generate_highlight_descriptione(instance_type):
   221         1          2.0      2.0     33.3      _instance_type = INSTANCE_TYPES.get(instance_type, None)
   222         1          1.0      1.0     16.7      if _instance_type is None:
   223                                                   # logging lookup failure on Prometheus
   224                                                   ec2_instance_lookup_failures.labels(reporter='API Events').inc()
   225                                                   return 'NA'
   226                                           
   227         1          1.0      1.0     16.7      cpu_type = _instance_type['extra']['physicalProcessor']
   228         1          0.0      0.0      0.0      num_vcpus = _instance_type['extra']['vcpu']
   229         1          0.0      0.0      0.0      ram_gb = _instance_type['extra']['memory']
   230         1          0.0      0.0      0.0      cloud_regions = _instance_type['extra']['regionCode']
   231         1          0.0      0.0      0.0      cloud = 'AWS'  # Update when multi cloud provider support is added
   232                                           
   233         1          1.0      1.0     16.7      description_text = f'{cpu_type} instance with {num_vcpus} vCPUs ' \
   234                                                                  f'and {ram_gb} of RAM, running on {cloud} ' \
   235                                                                  f'{cloud_regions} regions'
   236                                           
   237         1          1.0      1.0     16.7      return description_text
```
I'll push the code shortly, meanwhile please feel free to share your thoughts

ros/api/v1/hosts.py
```diff
@@ -390,29 +398,37 @@ class ExecutiveReportAPI(Resource):
...
+        "instance_types_highlights": fields.Nested(instance_types_highlights),
         "meta": fields.Nested(meta)
     }
 
     @marshal_with(report_fields)
     def get(self):
         org_id = org_id_from_identity_header(request)
-        system_queryset = system_ids_by_org_id(org_id, fetch_records=True)
-        systems_with_performance_record_queryset = db.session.query(PerformanceProfile.system_id)\
+        system_queryset = system_ids_by_org_id(org_id, fetch_records=True)  # Necesary for component state aggregations
```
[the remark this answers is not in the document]
saltgen: @kgaikwad Implemented usage of a cartesian product/anonymous table to use a single queryset and reduce DB calls (system_queryset was removed)

ros/lib/utils.py
```diff
@@ -125,6 +128,11 @@ def system_ids_by_org_id(org_id, fetch_records=False):
     return db.session.query(System.id).filter(System.tenant_id.in_(account_query))
 
 
+def performance_profiles_by_system_id(system_id, model):
```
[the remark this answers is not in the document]
saltgen: Util method removed

ros/lib/utils.py
```diff
@@ -185,3 +193,61 @@ def log_request(self, code='-', size='-'):
...
+    if instance_type not in INSTANCE_TYPES.keys():
+        # logging lookup failure on Prometheus
+        ec2_instance_lookup_failures.labels(reporter='API Events').inc()
+        return 'NA'
+
+    cpu_type = INSTANCE_TYPES[instance_type]['extra']['physicalProcessor']
+    num_vcpus = INSTANCE_TYPES[instance_type]['extra']['vcpu']
+    ram_gb = INSTANCE_TYPES[instance_type]['extra']['memory']
+    cloud_regions = INSTANCE_TYPES[instance_type]['extra']['regionCode']
+    cloud = 'AWS'  # Update when multi cloud provider support is added
```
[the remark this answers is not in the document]
saltgen: Cloud provider will now be set as per the file type being looked up for instance type properties

ros/lib/utils.py
saltgen: Change pushed

ros/lib/utils.py
saltgen: Change pushed

saltgen:
Let's merge #275 after this one, thanks!

ros/api/v1/hosts.py
```diff
@@ -466,6 +500,29 @@ def get(self):
...
+        upload_date = (current_utc_datetime - timedelta(days=1)).date()
+
+        current_performance_profiles = [
+            record
+            for record in systems_with_performance_record_queryset
+            if record.report_date.date() == upload_date
+        ]
+
+        historical_performance_profiles = db.session.query(
+            PerformanceProfile).select_entity_from(
```
patilsuraj767, line 513: Instead of `select_entity_from()` can we use `aliased()`.  `select_entity_from()` is deprecated and will be removed from SQLAlchemy 2.0. 
Check the deprecation note - https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.select_entity_from

patilsuraj767 (changes requested):

ros/api/v1/hosts.py
saltgen: @patilsuraj767 Sure I will check, btw is there a way we can check these warnings?
Did this one just show up on the terminal?

saltgen: @patilsuraj767 Change pushed. I would like squash the commits before rebase and merge.

ros/api/v1/hosts.py
```diff
@@ -466,6 +501,31 @@ def get(self):
...
+
+        historical_performance_profiles = db.session.query(
+            systems_with_performance_record_subquery.c.system_id,
+            systems_with_performance_record_subquery.c.rule_hit_details
+        ).select_from(
+            aliased_performance_profile_history
+        ).join(
+            systems_with_performance_record_subquery,
+            systems_with_performance_record_subquery.c.system_id == aliased_performance_profile_history.system_id
+        )  # Systems older than 7 days/stale systems are considered here
```
patilsuraj767, line 523: @saltgen This is too complicated and also returning incorrect results.
If we just want to select the systems from `PerformanceProfileHistory` where `system_id = systems_with_performance_record_subquery.c.system_id`. Then please replace it with the below. 
```suggestion
       historical_performance_profiles = db.session.query(PerformanceProfileHistory).join(
            systems_with_performance_record_subquery,
            systems_with_performance_record_subquery.c.system_id == PerformanceProfileHistory.system_id
        ) # Systems older than 7 days/stale systems are considered here
```

patilsuraj767 (changes requested):
I am still reviewing the PR but for now please check the below.

ros/api/v1/hosts.py
```diff
@@ -466,6 +501,31 @@ def get(self):
         io_states_dict['undersized'] = -1
         total_conditions = totals['cpu'] + totals['memory'] + totals['io']
 
+        current_utc_datetime = datetime.now(timezone.utc)
+        upload_date = (current_utc_datetime - timedelta(days=1)).date()
+
+        current_performance_profiles = [
+            record
+            for record in systems_with_performance_record_queryset
+            if record.report_date.date() == upload_date
```
patilsuraj767, line 510: Sorry to asking again but I don't completely remember, Are we doing `date - 1` only to avoid the systems with state "Waiting for data"? Or Is there any other reason too?

saltgen: Np, we had run into a problem to select the date for the latest upload.

I was thinking this date could be dynamic but then there was no way to actually ascertain the correct latest upload date.

Hence, it was decided that we would use a static date i.e. yesterday's date.

ros/api/v1/hosts.py
saltgen: @patilsuraj767 Yeah this is cleaner I agree, just that I have added changes to select only certain columns for `PerformanceProfileHistory` so that entire records need not get loaded into memory.

Change pushed

ros/api/v1/hosts.py
saltgen: There is one more thing we could do is request a date from the client up until which he/she wants to see the report.

But, I think we will have to take that up in the next release.

saltgen: That said showing the report based on performance profiles only for one day feels a bit wrong, let me check with Pau

saltgen: @patilsuraj767 Change pushed, current profiles selection between today and stale_date

ros/api/v1/hosts.py
```diff
@@ -518,10 +576,16 @@ def get(self):
                     "under_pressure": cpu_states_dict['under_pressure']
                 }
             },
+            "instance_types_highlights": {
+                "current": highlights_instance_types(current_performance_profiles, 'current'),
+                "suggested": highlights_instance_types(current_performance_profiles, 'suggested'),
```
patilsuraj767, line 581: I think we should not use `current_performance_profiles` for suggested system. For suggested we should get the systems from `performance_profile_history` table.
Reason - If you check the executive report these systems are showed under `Most suggested instance types (yesterday)`. suppose user downloaded the report and he saw suggestion is to move to t2.medium from t2.small, then user added few more systems to ros say t2.large and t2.medium systems and downloaded the executive report again(on the same day). Now the section of `Most suggested instance types (yesterday)` will differ in earlier report and the new reported. Newly downloaded report will show new suggestion under **yesterday** section but hey system was created and reported today. 
So for suggested please select system from performance_profile_history table where report_date == current_date - 1.

patilsuraj767 (changes requested):

ros/api/v1/hosts.py
```diff
@@ -466,6 +500,30 @@ def get(self):
...
+        current_utc_datetime = datetime.now(timezone.utc)
+        stale_date = (current_utc_datetime - timedelta(days=7)).date()
+
+        current_performance_profiles = [
+            record
+            for record in systems_with_performance_record_queryset
+            if (
+                    record.report_date.date() == current_utc_datetime.date()
+                    and record.report_date.date() > stale_date
+            )
```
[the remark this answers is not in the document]
saltgen: @patilsuraj767 I have refactored the condition.

As per our discussion, `current_performance_profiles` are of systems between today(inclusive) and stale date(non-inclusive).

If we are explicit about this then even if a wrong `report_date` is reported by rules our report remains un-affected.

ros/api/v1/hosts.py
saltgen: As discussed, we would like to populate the `current` and  `suggested` section from the same rule hit to show the latest of suggestions per performance_profile.

The same don't exist in the historical table atm.

patilsuraj767: resolving this conversation considering we need to remove **yesterday** keyword from the executive report frontend.

patilsuraj767 (approved):
/lgtm 
Thanks for the PR @saltgen

patilsuraj767:
Thanks @PreetiW for testing the PR in ephemeral environment.

PR #253, opened 2022-08-11
Refs #RHIROS-731 - Increase SQLAlchemy DB pool size

patilsuraj767:
Seems like API caching has not solved the QueuePool size limit issue. We will need to increase the `SQLALCHEMY_POOL_SIZE`. 

I have checked other services like [vulnerability-engine](https://github.com/RedHatInsights/vulnerability-engine/blob/d31b67a7223ff69ee3c4220e79304268f2363c62/deploy/clowdapp.yaml#L914), [payload-tracker](https://github.com/RedHatInsights/payload-tracker/blob/827f66ce3c2810abcfc3acfa9cee1021141827ad/deploy/deploy.yml#L259) etc they are having higher pool size.

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -59,6 +59,10 @@ objects:
             value: ${CLOWDER_ENABLED}
           - name: ENABLE_RBAC
             value: "${ENABLE_RBAC}"
+          - name: DB_POOL_SIZE
+            value: ${DB_POOL_SIZE}
+          - name: DB_MAX_OVERFLOW
+            value: ${DB_MAX_OVERFLOW}
     - name: processor
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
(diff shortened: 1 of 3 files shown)


r14chandra (approved):
looks good 👍🏻

clowdapp.yaml
```diff
@@ -202,3 +210,7 @@ parameters:
   value: "localhost"
 - name: POPULATOR_LOG_FORMAT
   value: cloudwatch 
+- name: DB_POOL_SIZE
+  value: "20"
+- name: DB_MAX_OVERFLOW
+  value: "50"
```
kgaikwad, line 216: Instead of setting `max_overflow` directly to `50`, could you please set it to `10 or 20` in the beginning? 
If anyone would like to change it then one can override the parameter, correct?

patilsuraj767: ok, I will change it to 20.

kgaikwad (approved):

PR #254, opened 2022-08-11
RHIROS-781 optimize /is_configured db queries

r14chandra:

PR #255, opened 2022-08-17
RHIROS-783 fix db query for /call_to_action

r14chandra:


ros/api/v1/call_to_action.py
```diff
@@ -1,9 +1,9 @@
 from flask_restful import Resource
 from flask import request
 from ros.lib.utils import (
-    org_id_from_identity_header, system_ids_by_org_id)
+    org_id_from_identity_header, systems_with_profile)
```
kgaikwad, line 4: @r14chandra - Any specific reason that you extracted query part into a separate method?

r14chandra: We can use this method in the PR for /is_configured - #254. Once this PR is in, I'll make the required changes in that PR.

r14chandra: No. of systems having performance profile is used in some place like under `is_configured` and `executive_report`

ros/lib/utils.py
```diff
@@ -158,6 +158,11 @@ def calculate_percentage(numerator, denominator):
         return 0
 
 
+def systems_with_profile(org_id):
```
kgaikwad, line 161: Will it make sense to include one more function argument which will accept the extra filters specific to context?

kgaikwad, line 161:
```suggestion
def systems_ids_with_profile(org_id):
```

This method returns system ids and not system/profile records. Could you rename this method to something more specific?

kgaikwad (changes requested):

ros/lib/utils.py
r14chandra: I've only seen filter for `number_of_recommendations`. Not generic. Is it still needed?

kgaikwad: > I've only seen filter for `number_of_recommendations`. Not generic. Is it still needed?

Then, it won't be needed. One query - do we really required `number_of_recommendations` filter? In which scenario system will have a performance profile record without any recommendation?

r14chandra: `Optimized` state archive.

kgaikwad (approved):
LGTM 👍

kgaikwad:
Thanks @r14chandra!

PR #256, opened 2022-08-23
RHIROS-733 | Cloud provider check for Update events

saltgen:
Changes:
1. Added an assert for cloud provider value for  update events
2. Added unittest

Additional:
- NA

### DIFF
--- a/ros/processor/inventory_events_consumer.py
+++ b/ros/processor/inventory_events_consumer.py
@@ -131,6 +131,7 @@ class InventoryEventsConsumer:
         if (
                 msg['platform_metadata'] is None
                 and msg['type'] == 'updated'
+                and msg['host']['system_profile']['cloud_provider']
         ) or 'is_ros' in msg['platform_metadata']:
             LOG.info(
                 '%s - Processing a message for system(%s) belonging to account: %s and org_id: %s',
--- a/tests/test_inventory_events_consumer.py
+++ b/tests/test_inventory_events_consumer.py
@@ -12,7 +12,7 @@ PERFORMANCE_RECORD = {'total_cpus': 1, 'instance_type': 't2.micro', 'mem.physmem
                       'mem.util.free': 161906.595, 'region': 'ap-south-1'}
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
 def inventory_event_message():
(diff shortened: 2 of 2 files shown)


ros/processor/inventory_events_consumer.py
```diff
@@ -131,6 +131,7 @@ def host_create_update_events(self, msg):
         if (
                 msg['platform_metadata'] is None
                 and msg['type'] == 'updated'
+                and msg['host']['system_profile']['cloud_provider']
         ) or 'is_ros' in msg['platform_metadata']:
```
patilsuraj767, line 135: Due to this "or" condition there can be one edge case in which hosts(VM, baremetal) will get processed. 
If user had run ros ansible playbook on baremetal host than puptoo will set `is_ros=True` for that host's insights report and ros-backend will process it.
Might be I am overthinking but possible.

saltgen: @patilsuraj767 In that case wouldn't that be a putpoo thing to sort out?

patilsuraj767: Yes, this can be fixed from puptoo side.
Approving the PR. Thanks @saltgen

patilsuraj767: One more thing, I remembered in the past @kgaikwad mentioned for few **update** message like update of display_name full host object is not passed in kafka topic. In that case this `msg['host']['system_profile']['cloud_provider']` will return null.

saltgen: Checking, thanks @patilsuraj767

saltgen: Couldn't look into this issue yet, cause of an existng issue with host-inventory, would be great if someone can help me with the verification and changes if any, thanks!

patilsuraj767 (approved):
I have verified the display_name update event from inventory it do sent the cloud_provider details. http://pastebin.test.redhat.com/1077393

I am approving the PR. @kgaikwad @saltgen feel free to merge it whenever we need to push this to stage.

saltgen:
> I have verified the display_name update event from inventory it do sent the cloud_provider details. http://pastebin.test.redhat.com/1077393
> 
> I am approving the PR. @kgaikwad @saltgen feel free to merge it whenever we need to push this to stage.

Thanks! @patilsuraj767 :rocket:

PR #257, opened 2022-09-08
RHIROS-793 event producer for notifications

r14chandra:


ros/lib/consume.py
```diff
@@ -0,0 +1,16 @@
+from confluent_kafka import Consumer
+from ros.lib.config import INSIGHTS_KAFKA_ADDRESS, GROUP_ID, kafka_auth_config
+
+
+def init_consumer(kafka_topic):
```
kgaikwad, line 5: @r14chandra  - could you please create a separate PR for this refactoring with consume & produce file?

kgaikwad (changes requested):
Thank you @r14chandra! 
Added few inline comments. Tests are not happy, you will need to handle none case.

ros/processor/event_producer.py
```diff
@@ -0,0 +1,62 @@
...
+from ros.lib.config import NOTIFICATIONS_TOPIC, get_logger
+
+logger = get_logger(__name__)
+
+
+# Event for new recommendation
+def new_recommendation_event(host, platform_metadata):
+    request_id = platform_metadata.get('request_id')
+    payload = {
+        "version": "v1.0.0",
```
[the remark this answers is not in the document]
r14chandra: You can find it here - [notification_payload](https://consoledot.pages.redhat.com/notifications-docs/dev/user-guide/send-notification.html#_kafka)

ros/processor/insights_engine_result_consumer.py
```diff
@@ -189,6 +197,14 @@ def process_report(self, host, platform_metadata, reports, system_metadata, perf
...
                 )
                 db.session.commit()
+
+                # Trigger event for notification
+                if system_previous_state not in (SYSTEM_STATES['OPTIMIZED'], system_current_state):
+                    LOG.info(
+                        f"{self.prefix} - New recommendation event triggered for the system: {host['id']}"
+                    )
+                    new_recommendation_event(host, platform_metadata)
+
```
kgaikwad, line 207: Method `process_report` is getting bigger. All engine logic is mainly kept under `process_report` method. For better readability & reusability, could you try to extract out this notification part into a separate small method?

ros/processor/event_producer.py
```diff
@@ -0,0 +1,62 @@
...
+        "bundle": "rhel",
+        "application": "ros",
+        "event_type": "",
```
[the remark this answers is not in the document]
r14chandra: It is required. I left it to confirm once. I'll add.

ros/processor/insights_engine_result_consumer.py
```diff
@@ -189,6 +197,14 @@ def process_report(self, host, platform_metadata, reports, system_metadata, perf
                     f"{self.prefix} - Performance profile created/updated successfully for the system: {host['id']}"
                 )
                 db.session.commit()
+
+                # Trigger event for notification
+                if system_previous_state not in (SYSTEM_STATES['OPTIMIZED'], system_current_state):
```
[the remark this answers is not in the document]
r14chandra: I need to replace previous state with current state

ros/processor/event_producer.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+    except KafkaError:
+        logger.exception(
+            "Failed to produce message to [%s] topic: %s", NOTIFICATIONS_TOPIC, request_id
+        )
+
+
+def upload_message_to_notification(payload, request_id):
+    bytes_ = json.dumps(payload).encode('utf-8')
+    producer = produce.init_producer()
```
kgaikwad, line 74: Does it make sense to create a new kafka producer instance on every new notification?
I think you should use a single kafka producer instance to publish these messages.

ros/processor/event_producer.py
```diff
@@ -0,0 +1,76 @@
...
+from ros.lib.config import NOTIFICATIONS_TOPIC, get_logger
+from ros.lib.utils import org_id_from_identity_header, systems_ids_for_existing_profiles
+
+logger = get_logger(__name__)
+
+
+# Event for new suggestion
+def new_suggestion_event(host, platform_metadata, previous_state, current_state):
+
+    org_id = org_id_from_identity_header(request)
```
kgaikwad, line 16: IIUC for existing processors, we are fetching org_id from either host object or platform_metadata.
I didn't get this part. Is request object present for background processing?
Why can't you use `host` object here to fetch `org_id` same way you did here -> [line](https://github.com/RedHatInsights/ros-backend/pull/257/files#diff-7f50c9ee82de30d0024864fef8fd7f6b8245e30a4ae9058582844eb544a52156R27)

ros/processor/event_producer.py
```diff
@@ -0,0 +1,76 @@
...
+    query = systems_ids_for_existing_profiles(org_id)
+    systems_with_suggestions = query.filter(PerformanceProfile.number_of_recommendations > 0).count()
+    request_id = platform_metadata.get('request_id')
+    payload = {
```
kgaikwad, line 20: Could you extract keys that commonly used for every notification example - `version, bundle, application`?
Then just append those keys which are specific to current event. What do you think?

ros/processor/insights_engine_result_consumer.py
```diff
@@ -200,3 +214,15 @@ def process_report(self, host, platform_metadata, reports, system_metadata, perf
...
+    def trigger_notification(
+        self, system, account, host, platform_metadata, system_previous_state, system_current_state
+    ):
+        if system_previous_state:
+            if system_current_state not in (SYSTEM_STATES['OPTIMIZED'], system_previous_state[0]):
+                LOG.info(
+                    "%s - New suggestion event triggered for the system: %s belonging" +
+                    "to account: %s (%s) and org_id: %s",
+                    self.prefix, system.inventory_id, account.account, account.id, account.org_id
+                )
```
kgaikwad, line 227:
```suggestion
                LOG.info(
                    "%s - Triggering a new suggestion event for the system: %s belonging" +
                    "to account: %s (%s) and org_id: %s",
                    self.prefix, system.inventory_id, account.account, account.id, account.org_id
                )
```
At this point, message is not yet triggered. You can either move this log statement below the `new_suggestion_event` method call or modify this log statement.

I would prefer to log this inside a event_producer where it actually triggers the message.

kgaikwad (changes requested):

ros/processor/event_producer.py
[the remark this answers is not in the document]
r14chandra: > In above shared document, it says current version is "v1.2.0". Any specific reason that we are not using latest one?

As confirmed by them, this field is not important now. I'll remove it along with `rh-message-id` (as this should be under kafka header. Again this is not required again to send).

ros/processor/event_producer.py
r14chandra: Sorry I didn't get it.

ros/processor/event_producer.py
```diff
@@ -0,0 +1,76 @@
...
+        "application": "ros",
+        "event_type": "new-suggestion",
+        "timestamp": datetime.now(timezone.utc).isoformat(),
+        "account_id": host.get("account_id") or "",
+        "org_id": host.get("org_id"),
+        "context": {
+            "event_name": "New suggestion",
+            "systems_with_suggestions": systems_with_suggestions
+        },
+        "rh-message-id": platform_metadata.get("request_id"),
```
gwenneg, line 32: This line will cause a notification schema validation issue if you test the payload with [our validator](https://internal.console.stage.redhat.com/api/notifications/internal/index.html#/utils/message-validator):
```
$.rh-message-id: is not defined in the schema and the schema does not allow additional properties
```
The `rh-message-id` field is expected as a Kafka header but it's prefectly fine to omit it at first and add it later. If that header is present, your Kafka messages will be [protected against duplicate processing](https://core-platform-apps.pages.redhat.com/notifications-docs/dev/user-guide/send-notification.html#_protecting_your_kafka_messages_against_duplicate_processing).

gwenneg, line 32: This line will cause a notification schema validation error  if you test the payload with our validator:
```
$.rh-message-id: is not defined in the schema and the schema does not allow additional properties
```
The `rh-message-id` field is expected as a Kafka header but it's prefectly fine to omit it at first and add it later. If that header is present, your Kafka messages will be protected against duplicate processing.

ros/processor/event_producer.py
r14chandra: Changes pushed 👍🏻

ros/processor/event_producer.py
```diff
@@ -0,0 +1,71 @@
...
+logger = get_logger(__name__)
+
+
+# Event for new suggestion
+def new_suggestion_event(host, platform_metadata, previous_state, current_state, producer):
+
+    org_id = host.get("org_id")
+    query = systems_ids_for_existing_profiles(org_id)
+    systems_with_suggestions = query.filter(PerformanceProfile.number_of_recommendations > 0).count()
+    request_id = platform_metadata.get('request_id')
```
kgaikwad, line 17: I would like to confirm that does engine message always include `platform_metadata` info?

ros/processor/insights_engine_result_consumer.py
```diff
@@ -200,3 +221,16 @@ def process_report(self, host, platform_metadata, reports, system_metadata, perf
                 ).inc()
                 LOG.error("%s - Unable to add system %s to DB belonging to account: %s and org_id: %s - %s",
                           self.prefix, host['id'], account.account, account.org_id, err)
+
+    def trigger_notification(
+        self, system, account, host, platform_metadata, system_previous_state, system_current_state
+    ):
+        print("** ", system_previous_state, system_current_state)
```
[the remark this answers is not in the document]
r14chandra: Oops 😓

ros/processor/event_producer.py
r14chandra: Yes,  it does

kgaikwad (approved):
Though this PR requires some more refactoring around event producer, I am OK to go ahead with this as first version of notification.

kgaikwad:
Thanks @r14chandra :ok_hand::tada::ice_cream: for addressing all comments & driving end to end notification integration

PR #258, opened 2022-09-09
add consume and produce files

r14chandra:


kgaikwad (approved):
Legit :+1: 
Thank you @r14chandra for extracting out initialization part for consume & produce into separate files.

PR #260, opened 2022-09-22
RHIROS-777 | Add non PSI system count in Exec Report response

saltgen:
### Changes:

1. DB Schema update to save psi_enabled data to performance profile
2. Changes added for non_psi_count for exec report
3. Updated openapi.json
4. Added couple of unittests

### Additional:
NA


migrations/versions/25e44a14ecb8_added_psi_enabled.py
```diff
@@ -0,0 +1,34 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('performance_profile', 'report_date',
+               existing_type=postgresql.TIMESTAMP(timezone=True),
+               nullable=True)
+    op.add_column('systems', sa.Column('psi_enabled', sa.Boolean(), nullable=True))
+    # ### end Alembic commands ###
```
kgaikwad, line 25: As a thumb rule, both tables `performance_profile` & `performance_profile_history` should be identical in terms of schema. Could you please make change in `performance_profile_history` table as well?

migrations/versions/25e44a14ecb8_added_psi_enabled.py
```diff
@@ -0,0 +1,34 @@
...
+# revision identifiers, used by Alembic.
+revision = '25e44a14ecb8'
+down_revision = '290a91f4eff6'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('performance_profile', 'report_date',
```
kgaikwad, line 21: Why this is needed for PSI enabled change?

kgaikwad (changes requested):

migrations/versions/25e44a14ecb8_added_psi_enabled.py
saltgen: This is auto-generated, shall I remove the upgrade func.?

migrations/versions/25e44a14ecb8_added_psi_enabled.py
saltgen: I have added the field on Systems, do we need this info. on `performance_profile` and `performance_profile_history`?

saltgen: @kgaikwad Lets check with Pau if we would like to include `psi_enabled` in profile tables as well, thanks!

migrations/versions/25e44a14ecb8_added_psi_enabled.py
kgaikwad: Did you mean below line is auto-generated in the migration?

```
op.alter_column('performance_profile', 'report_date',
               existing_type=postgresql.TIMESTAMP(timezone=True),
               nullable=True)
```

migrations/versions/25e44a14ecb8_added_psi_enabled.py
pgarciaq: @saltgen I don't really know if we would like to include `psi_enabled` or not. What do we get/lose by adding/not adding it?

saltgen: Adding this as a discussion point for today's Updates call

migrations/versions/25e44a14ecb8_added_psi_enabled.py
saltgen: @kgaikwad Yes, that is correct, was present when migration file was generated.

migrations/versions/25e44a14ecb8_added_psi_enabled.py
saltgen: @pgarciaq We have these fields available on every `performance_profile`.

```
performance_record
performance_utilization
report_date
state
operating_system
rule_hit_details
number_of_recommendations
system_id
```
The same record gets copied over to `performance_profile_history` table. 

We do this to keep track of the system changes over time.

W.r.t above do you think `psi_enabled` is something we need to keep track of over the lifetime of a system?

migrations/versions/25e44a14ecb8_added_psi_enabled.py
saltgen: @patilsuraj767 They have added a disclaimer as well, he he
```
def upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
```
I'll check and refactor, thanks.

migrations/versions/25e44a14ecb8_added_psi_enabled.py
saltgen: As discussed, moving ahead with `psi_enabled` on `performance_profile` and `performance_profile_history` tables, respectively.

saltgen: @kgaikwad Changes pushed

migrations/versions/25e44a14ecb8_added_psi_enabled.py
saltgen: @kgaikwad @patilsuraj767 There is a new migration file now, changes pushed

migrations/versions/3fb61393a227_added_psi_enabled.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = '3fb61393a227'
+down_revision = '290a91f4eff6'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('performance_profile', sa.Column('psi_enabled', sa.Boolean(), nullable=True))
```
kgaikwad, line 21: @saltgen  -
psi_enabled will be calculated based on availability of pressure metrics data. Will it make sense to keep default value as `False` and column as non-nullable?  
column value becomes True only when pressure metrics data available.

ros/lib/models.py
```diff
@@ -10,12 +10,16 @@ class PerformanceProfile(db.Model):
     performance_record = db.Column(JSONB)
     performance_utilization = db.Column(JSONB)
     report_date = db.Column(
-        db.DateTime(timezone=True), default=datetime.datetime.utcnow)
+        db.DateTime(timezone=True),
+        default=datetime.datetime.utcnow,
+        nullable=False
+    )
```
kgaikwad, line 16: In my opinion, this change should go separately as it is not dependent on psi-enabled. This change is now blocked on this PR to get approved so to drive such common changes quickly, it would be good to create a separate PR. 
Currently, it's not blocking anyone's development so for now, it is OK.

kgaikwad (changes requested):

migrations/versions/3fb61393a227_added_psi_enabled.py
saltgen: @kgaikwad I took a look at all the non-nullable model fields and IMO `psi_enabled` doesn't qualify.
Also for some reason if we stop receiving value for the field or any respective record does not get updated
the value will be set to `False`, I think this has the potential to cause confusion in future.

kgaikwad (approved):

kgaikwad:
Thank you @saltgen!

PR #264, opened 2022-09-28
Fixes #RHIROS-814 - Group histogram buckets based on url_rule instead…

patilsuraj767:
… of path


kgaikwad (approved):
Legit :+1:

kgaikwad:
Thank you @patilsuraj767 for fixing it.

PR #275, opened 2022-11-07
RHIROS-777 | Add PSI enabled system count in Exec Report response

saltgen:
Changes:
1. Addition of psi_enabled_count in API response

### DIFF
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -16,10 +16,12 @@ from ros.lib.models import (
     PerformanceProfileHistory,
 )
 from ros.lib.utils import (
-    is_valid_uuid, identity,
+    is_valid_uuid,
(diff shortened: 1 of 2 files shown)


PreetiW (approved):
Thank you @saltgen for the PR! 🙌🏻 
I have tested these changes in EE by uploading RHEL 7 (non-psi) and RHEL 8 (psi enabled) system and it is working as expected.

However we discussed and finalised that `non_psi_count` will not include RHEL 7 systems as there is no way to enable PSI on RHEL 7 and @saltgen is going to create another MR for same hence merging this PR!
cc: @kgaikwad 

image

PreetiW (approved):
@saltgen it will be great if you add a test for same in your next Ref PR. thanks 💯

PR #277, opened 2022-11-09
Get the processor and api ports from env vars if available

upadhyeammit:
I am unable to start the processor and api at the same time because both starts on port 8000. Should both listen on port 8000? if not then i will update that as well in this pr.


upadhyeammit:
To test:
export ROS_API_PORT=8002
pipenv run python3 -m ros.processor.main

and,
export ROS_API_PORT=8003
pipenv run python3 -m ros.api.main

patilsuraj767 (commented):
PR looks good. Can we also append `ROS_PROCESSOR_PORT=8004` to [scripts/.env](https://github.com/RedHatInsights/ros-backend/blob/main/scripts/.env) file so that we can directly run `. scripts/.env` to set all development variable at once instead of remembering variable name and running export command manually. 
Please use port 8004 as 8003 is used by puptoo in docker-compose.

saltgen:
@upadhyeammit Thanks for the changes!
@patilsuraj767 Do you think we can use something like [python-dotnev](https://pypi.org/project/python-dotenv/) as a dev package to _explicitly_ auto load the vars in the `.env` file

patilsuraj767:
@saltgen No, I think we should not add python-dotenv. 
Reason - 
1. We cannot add python-dotenv package as only dev dependency. 
2. To get python-dotenv to work we will need to edit main.py(which goes in production) and I think we should not touch the production code only to easy developer life from running `. .env` :sweat_smile:
3. It creates the risk of overriding env variables which are auto set by openshift/clowder in production.

upadhyeammit:
> PR looks good. Can we also append `ROS_PROCESSOR_PORT=8004` to [scripts/.env](https://github.com/RedHatInsights/ros-backend/blob/main/scripts/.env) file so that we can directly run `. scripts/.env` to set all development variable at once instead of remembering variable name and running export command manually. Please use port 8004 as 8003 is used by puptoo in docker-compose.

Done! Instead of 8002 and 8004 I have used 8004 and 8005

scripts/.env
```diff
@@ -1,2 +1,4 @@
 MINIO_ACCESS_KEY=BQA2GEXO711FVBVXDWKM
 MINIO_SECRET_KEY=uvgz3LCwWM3e400cDkQIH/y1Y4xgU4iV91CwFSPC
+ROS_PROCESSOR_PORT=8004
+ROS_API_PORT=8005
```
patilsuraj767, line 4: I think we should not add/change `ROS_API_PORT`, 
Reason -
1. If we change the `ROS_API_PORT` then we will also need to make change in the Makefile.
2. I think ros frontend by default use the port 8000 to connect to local backend while development, so that setup will so break, we will also need to make changes in frontend config. 

Better we only change `ROS_PROCESSOR_PORT` via `.env`.
```suggestion

```

saltgen:
> @saltgen No, I think we should not add python-dotenv. Reason -
> 
>     1. We cannot add python-dotenv package as only dev dependency.
> 
>     2. To get python-dotenv to work we will need to edit main.py(which goes in production) and I think we should not touch the production code only to easy developer life from running `. .env` sweat_smile
> 
>     3. It creates the risk of overriding env variables which are auto set by openshift/clowder in production.

```
We could list the package as a dev dependency in the Pipfile and then do a `pipenv install --dev`, but yes those will still need to be loaded in config.

Thanks for checking :v:

scripts/.env
upadhyeammit: The Makefile change is already is in this pr.
I am okay to keep the 8000 as default port however I would still want ROS_API_PORT to be configurable. 
So we have the flexibility to choose another port if there is any other services already listening on it and dont require to make change in code.

Considering this I will check in frontend how to depend on the env var for ROS_API_PORT

patilsuraj767 (approved):
Thanks for the PR @upadhyeammit. 
@saltgen feel free to merge once you are done reviewing. 
/lgtm

upadhyeammit:
Rebased!

saltgen:
@upadhyeammit Thanks for the changes! :confetti_ball:

PR #279, opened 2022-11-22
RHIROS-838 add tests for notification payload

r14chandra:


ros/processor/event_producer.py
```diff
@@ -8,12 +8,11 @@
 logger = get_logger(__name__)
 
 
-def new_suggestion_event(host, platform_metadata, previous_state, current_state, producer):
+def notification_payload(host, previous_state, current_state):
```
upadhyeammit, line 11: should we use same naming for parameters? as `system_previous_state` and `system_current_state`?

tests/test_insights_engine_result_consumer.py
```diff
@@ -216,3 +217,22 @@ def test_process_report_psi_enabled(engine_result_message, engine_consumer, db_s
...
+    platform_metadata = engine_result_message["input"]["platform_metadata"]
+    ros_reports = [engine_result_message["results"]["reports"][7]]
+    system_metadata = engine_result_message["results"]["system"]["metadata"]
+    system_previous_state = "Idle"
+    engine_consumer.process_report(host, platform_metadata, ros_reports, system_metadata, performance_record)
+    system_record = db_get_host(host['id'])
+    response = notification_payload(
+                    host, system_previous_state, system_record.state)
+
+    assert response["account_id"] == host["account"]
```
upadhyeammit, line 234: Do you think we should also check below headers? With my limited knowledge of JS I can see reference of below payload headers in NotificationBackend,
~~~
        "bundle": "rhel",
        "application": "resource-optimization",
        "event_type": "new-suggestion",
        "context": {
            "event_name": "New suggestion",
            "systems_with_suggestions": systems_with_suggestions,
~~~

So if changing these will cause breakage in notification then I would suggest to assert on these as well? and if NotificationBackend depend on these then I vouch to make these as constants and remove these from method. 

https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/test/java/com/redhat/cloud/notifications/processors/email/aggregators/ResourceOptimizationPayloadAggregatorTest.java#L94-L96 
https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/test/java/com/redhat/cloud/notifications/processors/email/aggregators/ResourceOptimizationPayloadAggregatorTest.java#L100-L101

r14chandra: I don't think. The above fields' values (like `bundle` and all) won't change. I've added assert for the fields which are dynamic like `inventory_id`, `display_name`, and others.
Let me know if you have further queries or opinions on this.

ros/processor/event_producer.py
r14chandra: Done

tests/test_insights_engine_result_consumer.py
upadhyeammit: Exactly! If we are expecting these values to not change then we should have constants for them which makes it understandable to not touch them? and going one step ahead we should also have the unit test to confirm that those have not changed?

r14chandra: Done

upadhyeammit (approved):

PR #283, opened 2022-12-21
RHIROS-892 update os filtering logic

r14chandra:
Removed hardcoded os versions.
Now, we're showing os versions in the os filter that are registered with a user account.

- If all child versions are selected then their parent version will be selected automatically.
- If a parent version is selected then all its child versions will be selected automatically.

API docs
Screenshot from 2022-12-21 17-07-03

UI
Screenshot from 2022-12-21 17-05-41

### DIFF
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -189,12 +189,15 @@ class HostsApi(Resource):
                 # FIXME: remove `if` block after getting OS values as numeric from frontend
                 if "RHEL" in os:
                     os = os.split(" ")[1]
-                os_object = {"name": "RHEL"}
-                if len(os) == 1:
-                    os += ".0"
-                os_object["major"] = int(os.split(".")[0])
-                os_object["minor"] = int(os.split(".")[1])
-                modified_operating_systems.append(os_object)
+                if os.replace('.', '', 1).isdigit():
+                    os_object = {"name": "RHEL"}
+                    if len(os) == 1:
+                        os += ".0"
+                    os_object["major"] = int(os.split(".")[0])
+                    os_object["minor"] = int(os.split(".")[1])
+                    modified_operating_systems.append(os_object)
+                else:
+                    abort(400, message='Not a valid RHEL version')
             filters.append(System.operating_system.in_(modified_operating_systems))
         return filters
 
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -38,26 +38,6 @@ LOG = logging.getLogger(__name__)
 SYSTEM_STATES_EXCEPT_EMPTY = [
     "Oversized", "Undersized", "Idling", "Under pressure", "Storage rightsizing", "Optimized", "Waiting for data"
 ]
-OS_VERSIONS = {
-    '7.0': {"name": "RHEL", "major": 7, "minor": 0},
-    '7.1': {"name": "RHEL", "major": 7, "minor": 1},
-    '7.2': {"name": "RHEL", "major": 7, "minor": 2},
-    '7.3': {"name": "RHEL", "major": 7, "minor": 3},
-    '7.4': {"name": "RHEL", "major": 7, "minor": 4},
-    '7.5': {"name": "RHEL", "major": 7, "minor": 5},
-    '7.6': {"name": "RHEL", "major": 7, "minor": 6},
-    '7.7': {"name": "RHEL", "major": 7, "minor": 7},
-    '7.8': {"name": "RHEL", "major": 7, "minor": 8},
-    '7.9': {"name": "RHEL", "major": 7, "minor": 9},
-    '8.0': {"name": "RHEL", "major": 8, "minor": 0},
-    '8.1': {"name": "RHEL", "major": 8, "minor": 1},
-    '8.2': {"name": "RHEL", "major": 8, "minor": 2},
-    '8.3': {"name": "RHEL", "major": 8, "minor": 3},
-    '8.4': {"name": "RHEL", "major": 8, "minor": 4},
-    '8.5': {"name": "RHEL", "major": 8, "minor": 5},
-    '8.6': {"name": "RHEL", "major": 8, "minor": 6},
-    '9.0': {"name": "RHEL", "major": 9, "minor": 0},
-}
 
 SYSTEM_COLUMNS = [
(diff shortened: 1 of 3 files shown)


upadhyeammit:
@r14chandra when does OS version gets registered with user account?

r14chandra:
> @r14chandra when does OS version gets registered with user account?

when a user registers a system on c.r.c.

upadhyeammit:
> > @r14chandra when does OS version gets registered with user account?
> 
> when a user registers a system on c.r.c.

Should we care about showing stale versions available for filtering? I am assuming de-registering the system(s) does not remove the OS version? or does it?

r14chandra:
> > > @r14chandra when does OS version gets registered with user account?
> > 
> > 
> > when a user registers a system on c.r.c.
> 
> Should we care about showing stale versions available for filtering? I am assuming de-registering the system(s) does not remove the OS version? or does it?

It will not show that particular OS version.

ros/api/v1/hosts.py
```diff
@@ -206,10 +186,10 @@ def build_system_filters():
         if operating_systems := request.args.getlist('os'):
             modified_operating_systems = []
             for os in operating_systems:
-                os = os.split(" ")[1]
-                if os not in OS_VERSIONS.keys():
-                    abort(400, message='Not a valid RHEL version')
-                modified_operating_systems.append(OS_VERSIONS[os])
+                os_object = {"name": "RHEL"}
+                os_object["major"] = int(os.split(".")[0])
+                os_object["minor"] = int(os.split(".")[1])
```
upadhyeammit, line 191: By any chance either from UI or direct api call we will allow the call only for major version?
~~~
GET /api/ros/v1/systems?os=8
~~~

In this case we will get `IndexError: list index out of range` so asking?

r14chandra: Yes, we do. Here, in my logic `8.0` will work. I'll fix it.

ros/api/v1/hosts.py
```diff
@@ -206,10 +186,12 @@ def build_system_filters():
         if operating_systems := request.args.getlist('os'):
             modified_operating_systems = []
             for os in operating_systems:
-                os = os.split(" ")[1]
-                if os not in OS_VERSIONS.keys():
-                    abort(400, message='Not a valid RHEL version')
-                modified_operating_systems.append(OS_VERSIONS[os])
+                os_object = {"name": "RHEL"}
+                if len(os) == 1:
```
upadhyeammit, line 190: @r14chandra what if I want all the systems with major version `8` and I don't care about the minor versions? With this if condition we will return the systems which are `8.0` specifically? and I think below two queries are different,
1. Get all RHEL 8 systems
2. Get RHEL 8.0 systems

What do you think?

r14chandra: No, these would be treated as the same queries. Archives contain versions in `8.0` or `7.0` format not `8` or `7`.

PreetiW (approved):
Thank you for the PR @r14chandra , the changes looks good to me. I have tested the PR against the local frontend and everything works fine 💯 

attaching screenshots below for reference!

image

image

ros/api/v1/hosts.py
upadhyeammit: I was comparing the flow of UI with api and there frontend itself generates the major version filters upon getting minors and that confused me. So yeah, this looks good now from my side as well. Approving it!

upadhyeammit (approved):

ros/api/v1/hosts.py
```diff
@@ -206,10 +186,12 @@ def build_system_filters():
         if operating_systems := request.args.getlist('os'):
             modified_operating_systems = []
             for os in operating_systems:
-                os = os.split(" ")[1]
-                if os not in OS_VERSIONS.keys():
-                    abort(400, message='Not a valid RHEL version')
-                modified_operating_systems.append(OS_VERSIONS[os])
+                os_object = {"name": "RHEL"}
```
kgaikwad, line 189: Observation:
With this filter change, it will break the current flow once merged without frontend pull-request. 
@r14chandra, could you please add a code block which will make this functionality backward compatible with FIXME comment? So that once both side changes (frontend + backend) related to os filtering pushed to Production, one can delete this code block.

kgaikwad (changes requested):

ros/api/v1/hosts.py
kgaikwad: It can be addressed separately after discussing how we want to keep OS filtering via API going futher. 
Keeping this conversation open till our discussion.

ros/api/v1/hosts.py
r14chandra: Sure adding a fix.

ros/api/v1/hosts.py
```diff
@@ -206,10 +186,15 @@ def build_system_filters():
         if operating_systems := request.args.getlist('os'):
             modified_operating_systems = []
             for os in operating_systems:
-                os = os.split(" ")[1]
-                if os not in OS_VERSIONS.keys():
-                    abort(400, message='Not a valid RHEL version')
-                modified_operating_systems.append(OS_VERSIONS[os])
+                # FIX: Remove below if block once we get os values as numeric from frontend
```
kgaikwad, line 189: Nitpick - it would be good to use existing code tags i.e FIXME.

```suggestion
               # FIXME: remove `if` block after getting OS values as numberic by frontend
```

kgaikwad (commented):
@r14chandra - Except a small inline comment related to codetag, looks good.

kgaikwad:
@r14chandra - I have tested the changes locally, worked well with correct values in API.
But when I tried to send invalid values, it gave `500 INTERNAL SERVER ERROR`.
Example -  pass `os=RHEL+dsfdsf.dsfdsf` in API and see logs as below:

```

os_object["major"] = int(os.split(".")[0])
ValueError: invalid literal for int() with base 10: 'dsfdsf'
127.0.0.1 - - [03/Jan/2023:20:57:13 +0530] "GET /api/ros/v1/systems?os=9.0&os=RHEL+dsfdsf.dsfdsf HTTP/1.1" 500 37 "-" "curl/7.76.1"

```

kgaikwad (approved):
Thank you @r14chandra for adding quick fix to support backward compatibility.

PR #286, opened 2022-12-29
Add state to filter only if its valid

upadhyeammit:


ros/api/v1/hosts.py
```diff
@@ -197,9 +197,10 @@ def build_system_filters():
             modified_states = []
             for state in states:
                 state = state.capitalize()
-                modified_states.append(state)
                 if state not in SYSTEM_STATES_EXCEPT_EMPTY:
                     abort(400, message='values are not matching')
+                else:
+                    modified_states.append(state)
```
kgaikwad, line 203: nitpick - `else` not necessary 

```suggestion
                if state not in SYSTEM_STATES_EXCEPT_EMPTY:
                    abort(400, message='values are not matching')
                modified_states.append(state)
```

ros/api/v1/hosts.py
```diff
@@ -197,9 +197,10 @@ def build_system_filters():
             modified_states = []
             for state in states:
                 state = state.capitalize()
-                modified_states.append(state)
```
r14chandra, line 200: Previous and current changes both would work the same way. Is there any reason why we need this change?

saltgen: It is more of an optimization, looks like we can abort before an append operation

saltgen:
@upadhyeammit Would you like to add a couple of more refactors on this PR?

ros/api/v1/hosts.py
r14chandra: But append operation has constant time complexity.

upadhyeammit: I think we are doing unnecessary append to the list even if its invalid?

upadhyeammit:
> @upadhyeammit Would you like to add a couple of more refactors on this PR?

no as of now

kgaikwad (approved):
Thank you @upadhyeammit! Merging this pull-request.

kgaikwad:
Thank you @saltgen and @r14chandra for review.

PR #287, opened 2023-01-03
RHIROS-777 | Exclusion filter for RHEL 7 systems - non-psi-count

saltgen:
### Changes:
1. Added model field for anonymous table
2. Added RHEL 7 filter for non_psi_count
3. Added unit-test

### Additional:
'is' operator does not work well sqlalchemy equality comparisons. 
Hence, a #noqa comment as been used so that flake8 doesn't complain

Reference, [here](https://stackoverflow.com/questions/18998010/flake8-complains-on-boolean-comparison-in-filter-clause)

### DIFF
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -4,7 +4,7 @@ from sqlalchemy.types import Float
 from ros.lib.constants import SubStates
 
 from datetime import datetime, timedelta, timezone
-from sqlalchemy import asc, desc, nullslast, nullsfirst
+from sqlalchemy import asc, desc, nullslast, nullsfirst, Integer
 from flask_restful import Resource, abort, fields, marshal_with
 
 from ros.lib.models import (
@@ -438,7 +438,8 @@ class ExecutiveReportAPI(Resource):
             PerformanceProfile.system_id,
             PerformanceProfile.report_date,
             PerformanceProfile.rule_hit_details,
-            PerformanceProfile.psi_enabled
+            PerformanceProfile.psi_enabled,
+            PerformanceProfile.operating_system
         ).select_from(System).join(PerformanceProfile).filter(
             System.id == PerformanceProfile.system_id
             and System.id.in_(system_ids_by_org_id(org_id))
@@ -524,7 +525,11 @@ class ExecutiveReportAPI(Resource):
             PerformanceProfile.report_date <= stale_date
         ).count()
 
-        non_psi_count = systems_with_performance_record_queryset.filter_by(psi_enabled=False).count()
+        non_psi_count = systems_with_performance_record_queryset.filter(
+            PerformanceProfile.operating_system['major'].astext.cast(Integer) != 7,
+            PerformanceProfile.psi_enabled == False  # noqa
+        ).count()
+
         psi_enabled_count = systems_with_performance_record_queryset.filter_by(psi_enabled=True).count()
 
         response = {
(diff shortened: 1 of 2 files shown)


ros/api/v1/hosts.py
```diff
@@ -524,7 +525,11 @@ def get(self):
             PerformanceProfile.report_date <= stale_date
         ).count()
 
-        non_psi_count = systems_with_performance_record_queryset.filter_by(psi_enabled=False).count()
+        non_psi_count = systems_with_performance_record_queryset.filter(
+            PerformanceProfile.operating_system['major'].astext.cast(Integer) != 7,
```
upadhyeammit, line 529: Does it make sense to keep psi_enabled False by default? and hence for RHEL 7 it will be False and we won't need this extra `!= 7` condition?

upadhyeammit, line 529: Does it make sense to keep psi_enabled False by default? and hence for RHEL 7 it will be False and we won't need this extra `!= 7` condition?

saltgen: Context:

I had the opinion that if we set any default value for the `psi_enabled` db column that would mean we are
assigning it with a value rather than consuming from rules, hence that field is default with `NULL`

W.r.t to the question, this is actually a count of such `psi_enabled=False` systems where-in we would like to !include RHEL 7 systems.

upadhyeammit: Thanks for explanation about the decision, continuing to that I am considering these,

1. Rules can give us value based on that we either make `psi_enabled` `true` or `false`
2. We don't have any queries where we check/depend if `psi_enabled` is `NULL`

I see above conditions are meeting so this means, 

a. We can actually make the `false` as default? 
b. So it will not matter what is the version of RHEL like 7 or 8? because for RHEL 7 also by default we will set `psi_enabled` as `false`?

saltgen: Yes, I agree `False` seems to be a better suited candidate. When we get a `True` for any RHEL 8 system we can just update that record.

Weird Scenario:

Let's say there is an issue on the Rules and it starts reporting `psi_enabled=True`, we will just end up saving that.

Also, this

```
>>> import this
The Zen of Python, by Tim Peters
...
...
Explicit is better than implicit.
...
...
```
Sorry if all this sounds too idealistic :sweat_smile: Please do share your thoughts :slightly_smiling_face:

upadhyeammit: If Rules engine is reporting `psi_enabled=True` then we need to refute those changes from ros side? I am yet to check if we already do it?

saltgen: AFAIK `psi_enabled` can be `False` or `True` for RHEL 8 systems i.e. PSI can be enabled/disabled;
Whereas enabling PSI is not a possibility for RHEL 7 systems, hence it's value is `False` by default.

This logic specific to RHEL 7 is not present in Rules atm, I will speak to Xiangce regarding this.

At the same time, IMO, ROS code should be tolerant of such changes because it depends quite a lot on
external data sources, largely data from Rules.

So to answer your question, I think, ROS code should store values that Rules report because it is our recommendation system, but have the logic as well to check if those values make sense before using the same.

PreetiW (approved):
Thank you @saltgen for the PR. I have tested it locally and everything looks fine 💯 
it is excluding systems with RHEL 7 for non_psi_count. Attaching screenshots for reference!

code wise everything looks good to me, LGTM 👍🏻  

image

image

image

image

ros/api/v1/hosts.py
upadhyeammit: We will have separate PR for this change, lets get this merged first!

upadhyeammit (approved):

PR #288, opened 2023-01-05
RHIROS-928| Exclusion filter for RHEL 7 systems - psi-enabled-count

saltgen:
### Changes:
1. Added RHEL 7 filter for psi-enabled-count

### DIFF
--- a/ros/api/v1/hosts.py
+++ b/ros/api/v1/hosts.py
@@ -4,7 +4,7 @@ from sqlalchemy.types import Float
 from ros.lib.constants import SubStates
 
 from datetime import datetime, timedelta, timezone
-from sqlalchemy import asc, desc, nullslast, nullsfirst, Integer
+from sqlalchemy import asc, desc, nullslast, nullsfirst
 from flask_restful import Resource, abort, fields, marshal_with
 
 from ros.lib.models import (
@@ -25,7 +25,7 @@ from ros.lib.utils import (
     count_per_state,
     calculate_percentage,
     org_id_from_identity_header,
-    highlights_instance_types,
+    highlights_instance_types, get_psi_count,
 )
 from ros.api.common.pagination import (
     limit_value,
@@ -515,12 +515,8 @@ class ExecutiveReportAPI(Resource):
             PerformanceProfile.report_date <= stale_date
         ).count()
 
-        non_psi_count = systems_with_performance_record_queryset.filter(
-            PerformanceProfile.operating_system['major'].astext.cast(Integer) != 7,
-            PerformanceProfile.psi_enabled == False  # noqa
-        ).count()
-
-        psi_enabled_count = systems_with_performance_record_queryset.filter_by(psi_enabled=True).count()
+        non_psi_count = get_psi_count(systems_with_performance_record_queryset, False)
+        psi_enabled_count = get_psi_count(systems_with_performance_record_queryset, True)
 
         response = {
             "systems_per_state": {
--- a/ros/lib/utils.py
+++ b/ros/lib/utils.py
@@ -8,6 +8,8 @@ import base64
 import json
 from flask import jsonify, make_response
 from flask_restful import abort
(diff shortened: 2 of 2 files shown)


saltgen:
@pgarciaq Hello Pau!
It looks like we might need a RHEL 7 filter for `psi_enabled_count` as well.
The reason for this is rules doesn't really have this _PSI cannot be enabled in RHEL 7_ logic. 

One fine day if there is a bug from rules and we end up saving a bunch of `psi_enabled=True` RHEL 7 records,
our psi numbers would look bad.

We can take this up with Xiangce but IMHO our code should be fault tolerant because it depends a lot on 
external sources of data.

Let me know your thoughts, thanks!

ros/api/v1/hosts.py
```diff
@@ -515,7 +515,11 @@ def get(self):
         ).count()
 
         non_psi_count = systems_with_performance_record_queryset.filter_by(psi_enabled=False).count()
-        psi_enabled_count = systems_with_performance_record_queryset.filter_by(psi_enabled=True).count()
+
+        psi_enabled_count = systems_with_performance_record_queryset.filter(
+            PerformanceProfile.operating_system['major'].astext.cast(Integer) != 7,
```
upadhyeammit, line 520: FYI change would benefit from : https://github.com/RedHatInsights/ros-backend/pull/287#discussion_r1062291169

saltgen: Ack :+1:

pgarciaq: > @pgarciaq Hello Pau! It looks like we might need a RHEL 7 filter for `psi_enabled_count` as well. The reason for this is rules doesn't really have this _PSI cannot be enabled in RHEL 7_ logic.
> 
> One fine day if there is a bug from rules and we end up saving a bunch of `psi_enabled=True` RHEL 7 records, our psi numbers would look bad.
> 
> We can take this up with Xiangce but IMHO our code should be fault tolerant because it depends a lot on external sources of data.
> 
> Let me know your thoughts, thanks!
> 
> P.S. non_psi_count RHEL 7 exclusion changes are already in review and will be merged. I'll wait on your ack just for this one.

+1

kgaikwad (changes requested):
@saltgen  - Could you please either make modification to existing [test case](https://github.com/RedHatInsights/ros-backend/blob/main/tests/test_api_endpoints.py#L279) or add new one?
Otherwise looks good to me!

PreetiW (changes requested):
Thank you for the PR @saltgen 💯 
I have tested the changes locally and it works fine, please see the screenshot attached for reference.

image

image

everything looks good from my side, only suggestion is if you can add a unit test for same then the PR will be good to merge. 👍🏻

saltgen:
@kgaikwad @PreetiW  I have wrapped the filtering code into a function cause,

1. code wasn't DRY
2. only one chunk of the code gets unit-tested now
3. needed to get rid of the hackish `#noqa` comment

The test suite takes 25 secs to run atm with 40 tests.
I can add another test for `True` input, but IMO it would be good to keep the test suite lean, whenever possible.

Thanks for testing Preeti :raised_hands: I don't think you'll need to retest :v:

PreetiW (approved):
@saltgen thanks for creating separate function for retrieving the count 👍🏻 LGTM 💯

kgaikwad:
> @kgaikwad @PreetiW I have wrapped the filtering code into a function cause,
> 
>     1. code wasn't DRY
> 
>     2. only one chunk of the code gets unit-tested now
> 
>     3. needed to get rid of the hackish `#noqa` comment
> 
> 
> The test suite already takes 25-28 secs to run with 40 tests. I can add another test for `True` input, but IMO it would be good to keep the test suite lean, whenever possible.
> 
As long as it checks with different data records then I guess it is fine to test one input (False/True).

kgaikwad (approved):

PR #289, opened 2023-01-06
PLMPGM-620 [SEC-APP-REQ-2] | CodeQL action for Static application security testing (SAST)

saltgen:
This change addresses the SEC-APP-REQ-2 SAST requirement as per ESSv9 assessment and recommendations for ROS

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,83 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'python' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Use only 'java' to analyze code written in Java, Kotlin or both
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,83 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
```
upadhyeammit, line 16: Do we need to mention the branch? As I understand we are delivering from `main` only, and older branches won't get any cherry picks?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,83 @@
...
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
```
upadhyeammit, line 19: Do we need to mention the branch? As I understand we are delivering from `main` only, and older branches won't get any cherry picks?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,83 @@
...
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+
+        # Details on CodeQL's query packs refer to : https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
+        # queries: security-extended,security-and-quality
+
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, Go, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
```
upadhyeammit, line 58: Do we need this for Python?

saltgen: Thanks, I agree, I don't think we need Autobuild, also it is an extra step.

I have commented it out and looks like the action is still passing

.github/workflows/codeql.yml
saltgen: As discussed, normally cherry picks should generally end up in the main branch and we don't have any other delivery/protected branches.
This would work without mentioning branch, I just thought we could go with it explicitly.

upadhyeammit: Explicit branch works as well! Thanks!

upadhyeammit:
Thanks @saltgen for PR, merging this now!

upadhyeammit (approved):

PR #300, opened 2023-03-01
RHIROS-456 fix inconsistent log format

r14chandra:

- [ ] Yes
- [X] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [X] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/ros/api/common/utils.py
+++ b/ros/api/common/utils.py
@@ -7,6 +7,7 @@ from ros.lib.utils import identity, user_data_from_identity, is_valid_uuid
 
 
 LOG = logging.getLogger(__name__)
+prefix = "VALIDATE REQUEST"
 
 
 def validate_rating_data(func):
@@ -64,10 +65,10 @@ def validate_rating_data(func):
         try:
             data = json.loads(request.data)
         except json.decoder.JSONDecodeError as err:
-            LOG.error('Decoding JSON has failed. %s', repr(err))
+            LOG.error(f"{prefix} - Decoding JSON has failed. {repr(err)}")
             abort(400, message="Decoding JSON has failed.")
         except TypeError as ex:
-            LOG.error('Invalid JSON format. %s', repr(ex))
+            LOG.error(f"{prefix} - Invalid JSON format. {repr(ex)}")
             abort(400, message="Invalid JSON format.")
 
         inventory_id = data['inventory_id']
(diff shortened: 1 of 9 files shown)


saltgen (commented):
Hi @r14chandra 
Is there any specific reason to move away from `fstrings`?

r14chandra:
> Hi @r14chandra Is there any specific reason to move away from `fstrings`?

I see we have used `f strings` at 4 places and the rest of the places use `% operator`. To make it consistent, I replaced it with % operator.

saltgen:
> > Hi @r14chandra Is there any specific reason to move away from `fstrings`?
> 
> I see we have used `f strings` at 4 places and the rest of the places use `% operator`. To make it consistent, I replaced it with % operator.

IIUC the format needs to be consistent. Here are a few reasons we should consider using `fstrings`,

- Readability
- Conciseness
- Better performance

r14chandra:
> IIUC the format needs to be consistent. Here are a few reasons we should consider using `fstrings`, 
>     * Readability
>     * Conciseness
>     * Better performance

So, should prefer using `fstring` in all of the places?

saltgen:
> > IIUC the format needs to be consistent. Here are a few reasons we should consider using `fstrings`,
> > * Readability
> > * Conciseness
> > * Better performance
> 
> So, should we prefer using `fstring` in all of the places?

That would be optimal, yes.

ros/api/common/utils.py
```diff
@@ -64,10 +64,10 @@ def validate_request(*args, **kwargs):
         try:
             data = json.loads(request.data)
         except json.decoder.JSONDecodeError as err:
-            LOG.error('Decoding JSON has failed. %s', repr(err))
+            LOG.error(f"Decoding JSON has failed. {repr(err)}")
             abort(400, message="Decoding JSON has failed.")
         except TypeError as ex:
-            LOG.error('Invalid JSON format. %s', repr(ex))
+            LOG.error(f"Invalid JSON format. {repr(ex)}")
```
saltgen, line 70: Shall we add prefixes for this validator function? 
Not a problem if it is static, like `{REQUEST-VALIDATOR}`.

Or you could check in *args/**kwargs if there is a class object or `__name__` and then 
`{type(a).__name__}-VALIDATOR`

I intend to overload this and use for future endpoints.

kgaikwad:
\retest

ShivamPandya:
@r14chandra  Tests have been passed now.

PR #301, opened 2023-03-09
RHIROS-987 - upgrade flask,werkzeug to newer versions

upadhyeammit:
* This updates the flask*, werkzeug and rest of the packages to latest version.
* This removes the pinning of the `markupsafe = "==2.0.1"` as I don't see any of the latest dependencies using `soft_unicode` from markupsafe. More about issue at https://github.com/aws/aws-sam-cli/issues/3661

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [x] Dependency Updated
- [ ] DB Migration Added


upadhyeammit:
This updates flask to 2.2.3, I am yet to go through change matrix of 2.2.1, 2.2.2 and 2.2.3 https://github.com/pallets/flask/releases

kgaikwad (changes requested):
@upadhyeammit, PR jobs are not happy. There is failure related to pytest.

upadhyeammit:
The sqlalchemy query api has been deprecated and we will need to migrate the queries as per new 2.0 specification : https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-20-query-usage . I will checking this further.

kgaikwad:
\retest

upadhyeammit:
After rebase I can see test failures :\ I am still working on to get all queries based on 2.0 spec

upadhyeammit:
As per the discussion with @yash2189 the EE failures should not be related to this change. He is checking further on this.

saltgen (changes requested):
@upadhyeammit Could you add some unittests for the code changes for example, the subquery change?
IMO it would be better to assert behavior consistency this way.

ros/api/v1/recommendations.py
```diff
@@ -44,14 +45,19 @@ class RecommendationsApi(Resource):
...
         if not is_valid_uuid(host_id):
             abort(404, message='Invalid host_id, Id should be in form of UUID4')
 
         ident = identity(request)['identity']
 
         filter_description = request.args.get('description')
 
-        system = system_ids_by_org_id(ident['org_id'], True).filter(System.inventory_id == host_id).first()
+        system_query = system_ids_by_org_id(ident['org_id'], True).filter(System.inventory_id == host_id)
+        try:
```
r14chandra, line 57: Do we require `try..except` block?
We have an `if` block at line 62 which is checking the system's existence.

ros/lib/utils.py
r14chandra, line 1: Line [41](https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py#L41) and 56 contain `session.query`, should it be replaced with `session.execute` or `session.scalars`?

ros/lib/utils.py
```diff
@@ -135,10 +134,11 @@ def sort_io_dict(performance_utilization: dict):
...
 
 def system_ids_by_org_id(org_id, fetch_records=False):
-    account_query = db.session.query(RhAccount.id).filter(RhAccount.org_id == org_id).subquery()
+    account_query = select(RhAccount.id).where(RhAccount.org_id == org_id)
+
     if fetch_records is True:
-        return db.session.query(System).filter(System.tenant_id.in_(account_query))
-    return db.session.query(System.id).filter(System.tenant_id.in_(account_query))
+        return select(System).filter(System.tenant_id.in_(account_query))
+    return select(System.id).filter(System.tenant_id.in_(account_query))
```
r14chandra, line 141: I didn't really get why we have replaced `db.session.query` with `select` in some places but not all.

upadhyeammit: If you see the code flow then `system_ids_by_org_id` is being used as input for other method as example https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py#L179 and if we revert this there sqlalchemy starts throwing warning about this type of passing db.session.query object wont be supported in future

ros/api/v1/recommendations.py
upadhyeammit: Only try..except is sufficient! I removed the if block!

ros/lib/utils.py
upadhyeammit: I moved those to scalars!

r14chandra: There are other places as well - like [recommendations.py#L59](https://github.com/RedHatInsights/ros-backend/blob/69bd289767c622b3ee5cc1a90c87998fd24feb1c/ros/api/v1/recommendations.py#L59), in garbage_collector file, tests folder.

upadhyeammit: Poor me! doing the changes for that as well!

upadhyeammit: Now there are very few queries where I did not find a way to not use .query,
~~~
$ grep 'db.session.query' ./* -R
./ros/api/v1/hosts.py:            db.session.query(PerformanceProfile, System, RhAccount)
./ros/api/v1/hosts.py:        systems_with_performance_record_queryset = db.session.query(
./ros/api/v1/hosts.py:        historical_performance_profiles = db.session.query(
./ros/lib/utils.py:    return db.session.query(PerformanceProfile.system_id) \
grep: ./scripts/minio-conf/certs: Permission denied
./tests/helpers/db_helper.py:    return db.session.query(model).filter_by(**filters)
~~~

apart from that I have moved all the queries to 2.0 way, also modifed the delete queries.

There are some questions or to do for me,
1. Look out if .query is being deprecated in future? I dont find it anywhere in docs
2. If yes then what and how we should form complex queries returning the models such as below using sqlalchemy select
~~~
-> count = query.count()
(Pdb) query.all()
[(<PerformanceProfile 86>, <System 86>, <RhAccount 86>)]
~~~

Request to review! thank you!

ros/processor/inventory_events_consumer.py
```diff
@@ -95,9 +96,11 @@ def host_delete_event(self, msg):
             LOG.debug(
                 f"{self.prefix} - Received a message for system with inventory_id {host_id}"
             )
-            rows_deleted = db.session.query(System.id).filter(System.inventory_id == host_id).delete()
+
+            rows_deleted = db.session.execute(db.delete(System).filter(System.inventory_id == host_id))
             db.session.commit()
-            if rows_deleted == 1:
+
+            if rows_deleted.rowcount > 0:
```
r14chandra, line 103: Is `rowcount` pre-defined anywhere?

And the second thing is, I see you have changed the condition from `==1` to `>0`. Any reason or difference in the result due to this check?

upadhyeammit: The rowcount is attribute on sqlalchemy object,

~~~
(Pdb) rows = db.session.execute(db.delete(System).filter(System.id == 1))
(Pdb) rows.rowcount
1
(Pdb) rows.rowcount
1
(Pdb) rows.rowcount
1
(Pdb) rows.__class__
<class 'sqlalchemy.engine.cursor.LegacyCursorResult'>
(Pdb) dir(rows)
['returns_rows', 'rowcount', 'scalar', 'scalar_one', 'scalar_one_or_none', 'scalars', 'supports_sane_multi_rowcount', 'supports_sane_rowcount', 'unique', 'yield_per']
(Pdb) rows = db.session.execute(db.delete(System).filter(System.id == 1))
(Pdb) rows.rowcount
0
~~~

and the below query can only delete one row considering inventory_id is unique and query has the '==' constraint?

~~~
            rows_deleted = db.session.execute(db.delete(System).filter(System.inventory_id == host_id))
~~~

So I think >0 is sane as well? however I can restore it back to '==0' if you think we should be very explicit about it.

ros/processor/inventory_events_consumer.py
```diff
@@ -135,7 +138,7 @@ def process_system_details(self, msg):
                 system = update_system_record(db.session, **system_fields)
                 if system is not None:
                     db.session.commit()
-                    account = db.session.query(RhAccount).filter_by(id=system.tenant_id).first()
+                    account = db.session.scalar(db.select(RhAccount).filter_by(id=system.tenant_id))
```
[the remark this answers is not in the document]
upadhyeammit: The difference is scalar vs scalars, and I think we can remove first from rest of the queries as well! Updating it

ros/lib/utils.py
```diff
@@ -151,11 +151,8 @@ def insert_performance_profiles(session, system_id, fields):
        performance_profile_history table.
     """
     fields = {} if fields is None else fields
-    old_profile_record = session.query(PerformanceProfile).filter_by(
-        system_id=system_id).first()
-    if old_profile_record:
-        session.delete(old_profile_record)
-        session.commit()
+    session.execute(db.delete(PerformanceProfile).filter(PerformanceProfile.system_id == system_id))
```
upadhyeammit, line 154: I think here we can just delete the PerformanceProfile instead of first checking if it exists? @r14chandra @kgaikwad ?

ros/processor/inventory_events_consumer.py
r14chandra: I think `==1` makes sense. Could you revert it back?

ros/lib/utils.py
r14chandra: Yeah,  we can do it. Just one thought, Does `commit`  have any consequences if we don't have anything to `execute`? As I see if we don't have anything to delete, the commit will always be executed. 
If not, then we can have this change.

upadhyeammit: As per my understanding running commit without deleting anything is not harmful. specially below para from the sqlalchemy(https://docs.sqlalchemy.org/en/20/orm/session_basics.html#committing),

> When there is no transaction in place for the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session), indicating that no operations were invoked on this [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) since the previous call to [Session.commit()](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.commit), the method will begin and commit an internal-only “logical” transaction, that does not normally affect the database unless pending flush changes were detected, but will still invoke event handlers and object expiration rules.

r14chandra (approved):
Looks good to me now 👍🏻. 
Thanks @upadhyeammit for resolving the comments!

ros/api/v1/recommendations.py
```diff
@@ -31,18 +31,19 @@ class RecommendationsApi(Resource):
 
     @marshal_with(data_fields)
     def get(self, host_id):
+        system = ''
```
kgaikwad, line 34: ditto
```suggestion
        system = None
```
`system` variable holds an object so does it make sense to initialize it to `None`?

ros/processor/insights_engine_consumer.py
```diff
@@ -93,10 +94,11 @@ def handle_msg(self, msg):
...
             )
-            reports = msg["results"]["reports"]  \
-                if msg["results"]["reports"] \
-                and type(msg["results"]["reports"]) == list \
-                else []
+            if msg["results"]["reports"] \
+                    and type(msg["results"]["reports"]) == list:
+                reports = msg["results"]["reports"]
+            else:
+                reports = []
```
kgaikwad, line 101: Nitpick - instead of `if..else` loop, does it make sense to initialize `reports` to empty array?

```suggestion
            reports = []
            if msg["results"]["reports"] \
                    and type(msg["results"]["reports"]) == list:
                reports = msg["results"]["reports"]
```

kgaikwad (commented):
Few suggestions added. Please feel free to address those in same or separate pull-request.

Pipfile
```diff
@@ -11,27 +11,26 @@ pytest-mock = "*"
...
 
 [packages]
-Flask = "~=1.1.2"
+Flask = "~=2.2.3"
 gunicorn = "~=20.0.4"
 flask-restful = "~=0.3.8"
-Flask-SQLAlchemy = "~=2.4.4"
-sqlalchemy = "<1.4"
+Flask-SQLAlchemy = "~=3.0.3"
+# anything >= 2.0.0 for sqlalchemy fails to auth with pgsql
```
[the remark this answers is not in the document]
upadhyeammit: You are correct! I fixed that by updating the env.py file, thanks for the find!

ros/api/v1/hosts.py
```diff
@@ -293,7 +298,7 @@ def get(self, host_id):
             RecommendationRating.rated_by == username
         ).first()
 
-        system = db.session.query(System).filter(System.inventory_id == host_id).first()
+        system = db.session.scalar(db.select(System).filter(System.inventory_id == host_id))
```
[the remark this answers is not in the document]
upadhyeammit: As discussed we are going to have separate pr for this change altogether

upadhyeammit:
> @upadhyeammit - jobs are not happy :)

I rebased, squashed and looks like we are green!

PR #304, opened 2023-03-16
Fixes #RHIROS-400 - fix improper authorization while submitting rating

kgaikwad:
## PR Title :boom:

Suggested formats: 

2. RHIROS-XXX Title 

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/ros/api/common/utils.py
+++ b/ros/api/common/utils.py
@@ -3,7 +3,12 @@ import logging
 from flask import request
 from flask_restful import abort
 from ros.lib.models import System, RatingChoicesEnum
-from ros.lib.utils import identity, user_data_from_identity, is_valid_uuid
+from ros.lib.utils import (
+    identity,
+    user_data_from_identity,
+    is_valid_uuid,
+    system_ids_by_org_id
+)
 
 
 LOG = logging.getLogger(__name__)
@@ -44,23 +49,22 @@ def validate_rating_data(func):
         if username is None:
             abort(403, message="Username doesn't exist")
 
-        return username
+        return ident, username
 
-    def check_for_system(inventory_id):
+    def check_for_system(ident, inventory_id):
         if not is_valid_uuid(inventory_id):
             abort(404, message='Invalid inventory_id, Id should be in form of UUID4')
 
-        system = System.query.filter(
-            System.inventory_id == inventory_id
-        ).first()
+        system_id = system_ids_by_org_id(ident['org_id']).filter(
+            System.inventory_id == inventory_id).first()
 
-        if system is None:
+        if system_id is None:
             abort(404, message=f"System {inventory_id} doesn't exist.")
 
-        return system.id
+        return system_id
 
     def validate_request(*args, **kwargs):
-        username = check_for_user()
+        ident, username = check_for_user()
         data = None
         try:
             data = json.loads(request.data)
@@ -72,7 +76,7 @@ def validate_rating_data(func):
             abort(400, message="Invalid JSON format.")
 
         inventory_id = data['inventory_id']
-        system_id = check_for_system(inventory_id)
+        system_id = check_for_system(ident, inventory_id)
         rating = check_for_rating(data)
         new_kwargs = {
             'rating': rating, 'username': username,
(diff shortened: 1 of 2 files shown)


ros/api/common/utils.py
```diff
@@ -44,23 +49,22 @@ def check_for_user():
...
 
-    def check_for_system(inventory_id):
+    def check_for_system(ident, inventory_id):
         if not is_valid_uuid(inventory_id):
             abort(404, message='Invalid inventory_id, Id should be in form of UUID4')
 
-        system = System.query.filter(
-            System.inventory_id == inventory_id
-        ).first()
+        system_id = system_ids_by_org_id(ident['org_id']).filter(
```
r14chandra, line 58: IMO, we can rate a recommendation for a system only when its performance profile is available. IIRC, we can see the system in ROS-UI when it has a performance profile. Is it good to check [systems_ids_for_existing_profiles](https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py#L166)?

upadhyeammit, line 58: nit, can we name the var to system_ids ?

upadhyeammit (commented):
the code looks good to me and I agree to @r14chandra suggestion on showing system only when performance profile is available, but maybe we should have different pr for that, maybe not? as the change will be small?

ros/api/common/utils.py
```diff
@@ -44,23 +49,22 @@ def check_for_user():
         if username is None:
             abort(403, message="Username doesn't exist")
 
-        return username
+        return ident, username
 
-    def check_for_system(inventory_id):
+    def check_for_system(ident, inventory_id):
```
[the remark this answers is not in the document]
kgaikwad: I agree with your concern.  Adding new params as per sorting order is bad idea.
My intention was not to sort but to add it as first argument as it is required + primary check in method logic. 
I will add tests around these validation methods.

ros/api/common/utils.py
kgaikwad: As I applied `first()` which returns single element, I think it is better to keep `system_id` as var. Thoughts?

kgaikwad: @r14chandra and @upadhyeammit  - On web UI, we also give option to rate for system when it is optimized with no suggestions. If you think we should allow this action ONLY when there will be suggestion(s) then as @upadhyeammit mentioned, could you please raise ticket and discuss with team? 
Thoughts?

r14chandra: Are you just considering the `optimized` state?
My concern is about rating a system (which is not there in UI) via post request from API not from UI.
And as you mentioned about this state, we can consider handling separately as it would also need the UI fix as well.

r14chandra (changes requested):

ros/api/common/utils.py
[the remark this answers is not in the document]
kgaikwad: Added test to verify validation check for authorization.

ros/api/common/utils.py
kgaikwad: > Are you just considering the optimized state?

@r14chandra - Where? In here, I am not considering it. I mentioned that current web ui allowing to add rating even when there is no recommendations i.e specific to when system is optimized.

r14chandra: > My concern is about rating a system (which is not there in UI) via post request from API not from UI.
This means we need to check for the performance profile table and not for the system table. 
Let me know if you didn't get it.

kgaikwad: @r14chandra  +1 for new change request. 
With this check, there will be inconsistency between web UI & API. 
As @upadhyeammit mentioned please create new change request, in my opinion we need to revise this feature completely.

r14chandra: Created a card -> [RHIROS-1012](https://issues.redhat.com/browse/RHIROS-1012)

r14chandra (approved):
LGTM

ros/api/common/utils.py
kgaikwad: Thanks!

upadhyeammit (approved):
Thanks for the test, LGTM!

PR #269, opened 2022-10-20
RHIROS-828 add inventory id to context

r14chandra:

PR #308, opened 2023-03-31
RHIROS-1035 add cloud_provider check to avoid processing on-premise archives

r14chandra:
To avoid processing archives that are not from cloud providers.

- [ ] Yes
- [X] No

## Security Checklist :lock:

- [X] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:
Merge this PR before - [insights-puptoo-#269](https://github.com/RedHatInsights/insights-puptoo/pull/269)

### DIFF
--- a/ros/lib/utils.py
+++ b/ros/lib/utils.py
@@ -78,21 +78,13 @@ def user_data_from_identity(identity):
     return identity['user']
 
 
-def validate_system(reporter, msg):
+def validate_ros_payload(is_ros, cloud_provider):
     """
-    Validate msg from consumers.
-    :param reporter: consumer reporter
-    :param msg: msg to validate.
-    :return: True if the msg is from a cloud provider and is_ros flag is set to true, False otherwise.
+    Validate ros payload.
+    :param is_ros: is_ros boolean flag
+    :param cloud_provider: cloud provider value
+    :return: True if cloud_provider is not none and is_ros flag is set to true, False otherwise.
     """
-    is_ros = False
-    cloud_provider = None
-    if reporter == 'INVENTORY EVENTS':
-        is_ros = msg["platform_metadata"].get("is_ros")
-        cloud_provider = msg['host']['system_profile'].get('cloud_provider')
-    else:
-        is_ros = msg["input"]["platform_metadata"].get("is_ros")
-        cloud_provider = msg["results"]["system"]["metadata"].get('cloud_provider')
     return True if is_ros and cloud_provider is not None else False
 
 
--- a/ros/lib/utils.py
+++ b/ros/lib/utils.py
@@ -1,5 +1,3 @@
-import ast as type_evaluation
-
 from collections import Counter
 from http.server import BaseHTTPRequestHandler
 import threading
@@ -80,20 +78,22 @@ def user_data_from_identity(identity):
     return identity['user']
 
 
-def validate_type(value, type_):
+def validate_system(reporter, msg):
     """
-    Validate the type of a value.
-    Currently available types: bool
-    :param value: Value to validate.
-    :param type_: Type to validate against.
-    :return: True if the value is of the specified type, False otherwise.
+    Validate msg from consumers.
+    :param reporter: consumer reporter
+    :param msg: msg to validate.
+    :return: True if the msg is from a cloud provider and is_ros flag is set to true, False otherwise.
     """
-    if type_ == bool:
-        # ast.literal_eval does not understand lowercase 'True' or 'False'
-        value = value.capitalize() if value in ['true', 'false'] else value
-    evaluated_value = type_evaluation.literal_eval(value) if value else None
-
-    return True if type(evaluated_value) == type_ else False
+    is_ros = False
+    cloud_provider = None
+    if reporter == 'INVENTORY EVENTS':
+        is_ros = msg["platform_metadata"].get("is_ros")
+        cloud_provider = msg['host']['system_profile'].get('cloud_provider')
+    else:
+        is_ros = msg["input"]["platform_metadata"].get("is_ros")
+        cloud_provider = msg["results"]["system"]["metadata"].get('cloud_provider')
+    return True if is_ros and cloud_provider is not None else False
 
 
 def cast_iops_as_float(iops_all_dict):
--- a/ros/processor/insights_engine_consumer.py
+++ b/ros/processor/insights_engine_consumer.py
@@ -13,7 +13,7 @@ from ros.lib.utils import (
     get_or_create,
     cast_iops_as_float,
     insert_performance_profiles,
-    validate_system
+    validate_ros_payload
 )
 from ros.processor.metrics import (processor_requests_success,
                                    processor_requests_failures,
@@ -83,7 +83,9 @@ class InsightsEngineConsumer:
                 self.consumer.commit()
 
     def handle_msg(self, msg):
-        is_valid = validate_system(self.reporter, msg)
+        is_ros = msg["input"]["platform_metadata"].get("is_ros")
+        cloud_provider = msg["results"]["system"]["metadata"].get('cloud_provider')
+        is_valid = validate_ros_payload(is_ros, cloud_provider)
         if is_valid:
             host = msg["input"]["host"]
             platform_metadata = msg["input"]["platform_metadata"]
--- a/ros/processor/insights_engine_consumer.py
+++ b/ros/processor/insights_engine_consumer.py
@@ -13,7 +13,7 @@ from ros.lib.utils import (
     get_or_create,
     cast_iops_as_float,
     insert_performance_profiles,
-    validate_type
+    validate_system
 )
 from ros.processor.metrics import (processor_requests_success,
                                    processor_requests_failures,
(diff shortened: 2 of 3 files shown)


ros/lib/utils.py
```diff
@@ -80,20 +78,22 @@ def user_data_from_identity(identity):
     return identity['user']
 
 
-def validate_type(value, type_):
+def validate_system(reporter, msg):
```
kgaikwad, line 81: If I understood correctly, by removing `validate_type` method, you are defeating the purpose of it.
With empty string in `is_ros`, this statement `is_ros and cloud_provider is not None` will return True.

ros/lib/utils.py
```diff
@@ -80,20 +78,22 @@ def user_data_from_identity(identity):
...
-
-    return True if type(evaluated_value) == type_ else False
+    is_ros = False
+    cloud_provider = None
+    if reporter == 'INVENTORY EVENTS':
+        is_ros = msg["platform_metadata"].get("is_ros")
+        cloud_provider = msg['host']['system_profile'].get('cloud_provider')
+    else:
+        is_ros = msg["input"]["platform_metadata"].get("is_ros")
+        cloud_provider = msg["results"]["system"]["metadata"].get('cloud_provider')
```
kgaikwad, line 95: Do you think we should capture info which helps in getting how many requests ros backend gets for non-cloud?

ros/lib/utils.py
kgaikwad, line 81:
```suggestion
def is_valid_payload(is_ros, cloud_provider):
```

This method needs  just`is_ros` & `cloud_provider` while validation.
Which objects/variables to pass as method arguments, it varies case to case basis and it's judgement call one can decide.

When either reporter's name changed or message object structure changes, one needs to touch this method. In my opinion, this method shouldn't care about getting `is_ros` & `cloud_provider` values . Instead of reporter and message, will it make sense to send only these two values directly.

kgaikwad (changes requested):
Thanks @r14chandra. Added inline comments.

ros/lib/utils.py
r14chandra: > how many requests ros backend gets for non-cloud?

IMO that is for future scope. It would be good to keep it simple as those logs would unnecessarily fill up our logs.

ros/lib/utils.py
upadhyeammit: Makes sense!

ros/lib/utils.py
upadhyeammit: Do you mean log statement or actually something app or org level to keep track of this? I think if we can keep track of org level maybe just using some var with log then also it will help as insight? However it wont persist on app reboot, maybe some table also to persist it just as count?

kgaikwad: > Do you mean log statement or actually something app or org level to keep track of this? I think if we can keep track of org level maybe just using some var with log then also it will help as insight? However it wont persist on app reboot, maybe some table also to persist it just as count?

yes, I would prefer to have short log statement.

ros/lib/utils.py
kgaikwad: > `is_ros` is not a string, it is a boolean flag. We only set this field when we have the output of `pmlog_summary` or `ros_config` parsers - [here](https://github.com/RedHatInsights/insights-puptoo/blob/master/src/puptoo/process/profile.py#L568).

Could you please confirm the value type of is_ros from a kafka messages that ros backend received? 
(Example - [here](https://github.com/RedHatInsights/ros-backend/blob/main/tests/data_files/events-message.json#L1152) which i am referring to)

r14chandra: > > `is_ros` is not a string, it is a boolean flag. We only set this field when we have the output of `pmlog_summary` or `ros_config` parsers - [here](https://github.com/RedHatInsights/insights-puptoo/blob/master/src/puptoo/process/profile.py#L568).
> 
> @r14chandra - Could you please confirm the value type of is_ros from a kafka messages that ros backend received? (Example - [here](https://github.com/RedHatInsights/ros-backend/blob/main/tests/data_files/events-message.json#L1152) which i am referring to)

We are adding `is_ros` as a boolean in [system_profile](https://github.com/RedHatInsights/insights-puptoo/blob/master/src/puptoo/process/profile.py#L569) but in platform_metadata, we are explicitly making it string - [here](https://github.com/RedHatInsights/insights-puptoo/blob/master/src/puptoo/app.py#L240-L241).
IMO we should consider making this field a boolean type to make it consistent.

@kgaikwad - To answer your query, the condition would anyway return false for the cases.

kgaikwad: > @kgaikwad - To answer your query, the condition would anyway return false for the cases.

My bad! I evaluated result using brackets i.e (is_ros and cloud_provide) 

@upadhyeammit @r14chandra  - I still think, we should prefer to use type casting (i.e `validate_type` on `is_ros`) on backend side. Thoughts?

upadhyeammit: I had discussion with @r14chandra and as per that its Puptoo which has inconsistency while generating the platform_metadata, so we should fix it and return True from there for is_ros. This means when we parse it back at ROS side it will be either True or None. So accordingly we can skip the typecasting.

kgaikwad: @r14chandra and @upadhyeammit,  +1 for fixing the value under platform_metadata on Puptoo side.
 
@r14chandra -  could you please raise the issue on puptoo side and mark this conversation resolve?

r14chandra: Here is the card - [ESSNTL-4611](https://issues.redhat.com/browse/ESSNTL-4611)

r14chandra: This PR is required to merge before [insights-puptoo-#269](https://github.com/RedHatInsights/insights-puptoo/pull/269)
as the current ROS stage contains [this](https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py#L93) conditions.

ros/lib/utils.py
upadhyeammit: > > I also discussed about this with the @r14chandra, and below are our thoughts,
> > ```
> > 1. We can have logs however those logs will only help for insights, and then Kibana instance is not that good when it comes to relating the logs from different services. I mean, something like, can I get total no of non cloud and is_ros systems when compared to central truth HBI? maybe not.
> > ```
> 
> Why we want to related logs from different services? Also, intention here is not to get total number of non-cloud systems vs is_ros systems.
> 
Yeah I was thinking to count total number of systems based on HBI.
> > ```
> > 2. The logs won't provide any advantage for troubleshooting? because we don't support this use case and even if customer complains why system is not in ROS then support team can actually check the facts in sosreport and communicate it.
> > ```
> 
> Definitely, this log statement will not be for troubleshooting.
> 
> > ```
> > 3. And yes we will fill up the kibana logs even when its not that useful.
> > ```
> 
> Could you suggest any alternative to get how many requests come to ROS backend where is_ros is true but not AWS instance? Though it's not useful for troubleshooting, we kept it for a use case. I guess you can confirm with Pau and decide to skip.

If there is use case then for sure we should add the logging, @pgarciaq request to confirm same.

ros/lib/utils.py
```diff
@@ -80,20 +78,14 @@ def user_data_from_identity(identity):
...
+    :param cloud_provider: cloud provider value
+    :return: True if cloud_provider is not none and is_ros flag is set to true, False otherwise.
     """
-    if type_ == bool:
-        # ast.literal_eval does not understand lowercase 'True' or 'False'
-        value = value.capitalize() if value in ['true', 'false'] else value
-    evaluated_value = type_evaluation.literal_eval(value) if value else None
-
-    return True if type(evaluated_value) == type_ else False
+    return True if is_ros and cloud_provider is not None else False
```
saltgen, line 88: @r14chandra IMO, it would be better to assert explicitly against possible values of `is_ros` and `cloud_provider`
Like, `if is_ros in [True, False]` and then assert `is_ros is True`, similarly for `cloud_provider`

saltgen (changes requested):
@r14chandra Could you modify the usage of `validate_payload` as a decorator like [so](https://github.com/RedHatInsights/ros-backend/blob/main/ros/api/v1/recommendation_ratings.py#L13)?

ros/lib/utils.py
```diff
@@ -80,20 +78,14 @@ def user_data_from_identity(identity):
     return identity['user']
 
 
-def validate_type(value, type_):
+def validate_payload(is_ros, cloud_provider):
```
saltgen, line 81:
```suggestion
def validate_ros_payload(is_ros, cloud_provider):
```

saltgen (changes requested):

ros/lib/utils.py
r14chandra: We just have one value for `is_ros` - True. That should be fine.

upadhyeammit (approved):

upadhyeammit:
Thank you @r14chandra for changes, thanks all for the reviews! I am merging this.
FYI: @r14chandra agreed to manage the release of Puptoo in sync with this change.

PR #316, opened 2023-04-21
RHIROS-1088 Cloud_provider check; avoid on-premise system processing

saltgen:

1. Necessary to avoid function invocation in case msg data validation fails
2. `is_ros` flag value is a string 'true' currently and it is difficult with existing code to verify integrity
3. Safer to assert `cloud_provider` value from an Enum of supported types
4. Brief logging in case any system processing is being ignored
5. unittests for engine and events consumer

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

Verification failure log:

Screenshot from 2023-04-21 23-44-27


ros/lib/utils.py
```diff
@@ -78,14 +79,43 @@ def user_data_from_identity(identity):
...
+        cloud_provider_flag = cloud_provider in [provider.value for provider in CloudProvider]
+
+        if (
+                is_ros_flag is True
+                and cloud_provider_flag is True
+        ):
+            return func(*args, **kwargs)
+        else:
+            LOG.warning(
+                f"{caller_prefix} - Validation failed - System {system_id}, "
```
kgaikwad, line 114: As this decorator will be called from same context, could you keep prefix consistent in this case? 
Reference Links:
https://github.com/RedHatInsights/ros-backend/blob/main/ros/processor/inventory_events_consumer.py#L29
https://github.com/RedHatInsights/ros-backend/blob/main/ros/processor/insights_engine_consumer.py#L42

kgaikwad (changes requested):
Thank you for addressing this change, added small inline comments. Please check and add your thoughts.

ros/lib/utils.py
```diff
@@ -78,14 +79,43 @@ def user_data_from_identity(identity):
...
+    def payload_validator(*args, **kwargs):
+        try:
+            payload_data = args[1]
+            if 'input' in payload_data.keys():  # input key is found only in engine msgs
+                caller_prefix = 'ENGINE CONSUMER'
+                account = payload_data['input']['host']['account']
+                org_id = payload_data['input']['platform_metadata']['org_id']
+                is_ros = payload_data["input"]["platform_metadata"].get("is_ros")
+                cloud_provider = payload_data["results"]["system"]["metadata"].get('cloud_provider')
+                system_id = payload_data["input"]["host"]["id"]
```
upadhyeammit, line 92: Can we have separate method for this code block?

ros/lib/utils.py
```diff
@@ -78,14 +79,43 @@ def user_data_from_identity(identity):
...
+            else:
+                caller_prefix = 'EVENTS CONSUMER'
+                account, org_id = payload_data['host']['account'], payload_data['host']['org_id']
+                system_id = payload_data['host']['id']
+                is_ros = payload_data["platform_metadata"].get("is_ros")
+                cloud_provider = payload_data['host']['system_profile'].get('cloud_provider')
```
upadhyeammit, line 98: and separate for this as well?

ros/lib/utils.py
```diff
@@ -78,14 +79,43 @@ def user_data_from_identity(identity):
...
+        except Exception as e:
+            LOG.error(e)
+            is_ros = False
+            cloud_provider, system_id, caller_prefix, account, org_id = None, None, None, None, None
+
+        is_ros_flag = True if is_ros in ['true', True, 'TRUE', '1'] else False
+        cloud_provider_flag = cloud_provider in [provider.value for provider in CloudProvider]
```
upadhyeammit, line 105: and separate for this as well?

upadhyeammit (commented):
I liked the refactoring, however if we can have meaningful small methods then it could help to improve readability of the code?

ros/lib/utils.py
```diff
@@ -78,14 +79,43 @@ def user_data_from_identity(identity):
...
+                account, org_id = payload_data['host']['account'], payload_data['host']['org_id']
+                system_id = payload_data['host']['id']
+                is_ros = payload_data["platform_metadata"].get("is_ros")
+                cloud_provider = payload_data['host']['system_profile'].get('cloud_provider')
+        except Exception as e:
+            LOG.error(e)
+            is_ros = False
+            cloud_provider, system_id, caller_prefix, account, org_id = None, None, None, None, None
+
+        is_ros_flag = True if is_ros in ['true', True, 'TRUE', '1'] else False
```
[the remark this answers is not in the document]
kgaikwad: >  I would prefer ROS to remain independent of such changes.

+1 to what you have mentioned @saltgen. But here, will it be good idea to add type check instead of possible values checking?  
Example - 
```
if is_ros & isinstance(is_ros, bool) :
    .. do something
    
```

In this way , ROS will get to know if any change in Puptoo or underneath library. Only thing is ROS won't process any archive file(s) till we address it inside ROS.
Thoughts?

kgaikwad (commented):
@saltgen  - As this PR includes multiple cards' fixes, could you please update commit message to include RHIROS-1048 along with RHIROS-1088 ? Something like - `Fixes #RHIROS-1088, #RHIROS-1048 - <message>`

saltgen:
Yet to add a test that cover non-ROS Inventory Update msg

ros/lib/utils.py
```diff
@@ -78,14 +88,53 @@ def user_data_from_identity(identity):
...
-    :param is_ros: is_ros boolean flag
-    :param cloud_provider: cloud provider value
-    :return: True if cloud_provider is not none and is_ros flag is set to true, False otherwise.
-    """
-    return True if is_ros and cloud_provider is not None else False
+def get_payload_metadata(payload_data):
+    caller_prefix = payload_data[0].prefix  # Fetching prefix from class instance
+    msg = payload_data[1]
+    try:
+        if caller_prefix == 'PROCESSING ENGINE RESULTS':
```
saltgen: @kgaikwad @upadhyeammit Does this logic look flaky to you?

ros/lib/utils.py
```diff
@@ -78,14 +90,53 @@ def user_data_from_identity(identity):
...
-    :param is_ros: is_ros boolean flag
-    :param cloud_provider: cloud provider value
-    :return: True if cloud_provider is not none and is_ros flag is set to true, False otherwise.
-    """
-    return True if is_ros and cloud_provider is not None else False
+def get_payload_metadata(payload_data):
+    caller_prefix = payload_data[0].prefix  # Fetching prefix from class instance
+    msg = payload_data[1]
+    try:
+        if caller_prefix == 'ENGINE RESULTS':
```
saltgen, line 97: @kgaikwad @upadhyeammit Does this condition logic look flaky to you?

ros/lib/utils.py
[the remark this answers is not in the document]
saltgen: Change pushed

ros/lib/utils.py
```diff
@@ -78,14 +79,43 @@ def user_data_from_identity(identity):
...
+                caller_prefix = 'ENGINE CONSUMER'
+                account = payload_data['input']['host']['account']
+                org_id = payload_data['input']['platform_metadata']['org_id']
+                is_ros = payload_data["input"]["platform_metadata"].get("is_ros")
+                cloud_provider = payload_data["results"]["system"]["metadata"].get('cloud_provider')
+                system_id = payload_data["input"]["host"]["id"]
+            else:
+                caller_prefix = 'EVENTS CONSUMER'
+                account, org_id = payload_data['host']['account'], payload_data['host']['org_id']
+                system_id = payload_data['host']['id']
```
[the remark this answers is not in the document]
saltgen: Done

ros/lib/utils.py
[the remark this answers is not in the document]
saltgen: Agreed, for now I am checking for `True` and `'true'`, but we can refactor this in future, change pushed

ros/lib/utils.py
saltgen: inheriting prefix from class instance, change pushed :+1:

ros/lib/utils.py
saltgen: This block resides inside `get_payload_metadata` now

ros/lib/utils.py
saltgen: I gave some thought about this one, post our discussion. IMO it would make sense for this native logic to remain directly inside the decorator, especially when it is not verbose.

saltgen:
### Processing of non-ROS archives

Screenshot from 2023-05-05 12-44-40

- The sample `non-ros-events-message.json` has been included in `tests/data_files`.
  The above message was generated by deleting following files from ROS archive, as per changes [here](https://github.com/RedHatInsights/insights-puptoo/blob/master/src/puptoo/process/profile.py#L568).

1. `insights.specs.Specs.ros_config.json`
2. `insights.specs.Specs.pmlog_summary.json`

**Note:**  Engine processor will not update system after failed Inventory update for non-ros archive, this is intentional.

ros/lib/utils.py
```diff
@@ -78,19 +90,58 @@ def user_data_from_identity(identity):
...
+            cloud_provider = msg["results"]["system"]["metadata"].get('cloud_provider')
+            account = msg['input']['host']['account']
+            org_id = msg['input']['platform_metadata']['org_id']
+        else:
+            inventory_id = msg['host']['id']
+            event_type = msg.get('type')
+            # exception for inventory update events
+            # is_ros flag is unexpected for the same
+            if event_type == 'updated':
+                is_ros = True
```
kgaikwad, line 109: I don't like the idea of setting `is_ros` flag here as at this level we are not sure whether payload related to system which already ROS enabled.

ros/processor/inventory_events_consumer.py
```diff
@@ -105,15 +112,15 @@ def host_delete_event(self, msg):
                     f"{self.prefix} - Deleted system with inventory id: {host_id}"
                 )
 
+    @validate_ros_payload
     def host_create_update_events(self, msg):
```
kgaikwad, line 116: For better understanding, will it be better to have a separate handling for update & create event type?

kgaikwad (changes requested):
Added few inline comments. Please let me know your thoughts.

ros/lib/utils.py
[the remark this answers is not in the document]
kgaikwad: I didn't get why you are checking for `true`, any specific reason?

ros/processor/inventory_events_consumer.py
```diff
@@ -135,17 +142,26 @@ def process_system_details(self, msg):
...
-                db.session.commit()
-                processor_requests_success.labels(
-                    reporter=self.reporter, org_id=account.org_id
-                ).inc()
-                LOG.info(
-                    f"{self.prefix} - Refreshed system {system.inventory_id} ({system.id}) "
-                    f"belonging to account: {account.account} ({account.id}) and org_id: {account.org_id}"
-                )
+                if msg['type'] == 'updated':
+                    system = update_system_record(db.session, System, **system_fields)
```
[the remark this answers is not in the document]
saltgen: This method can only _update_ existing records, if such record is found with inventory_id from system_fields.
Code, [here](https://github.com/RedHatInsights/ros-backend/pull/316/files#diff-73b796983793ca68d788f97a84b00ef19df5157d5fee7ee854c47ed1133d45f1R52-R56)

ros/lib/utils.py
saltgen: My 2 cents:

1. As mentioned above, any ROS system record will only be updated if found with `update_system_record` method
2. This is temporary, we should ideally get system_metadata in kafka msg, once that is fixed we can remove this if block

So, I think this should be fine, for now.

ros/processor/inventory_events_consumer.py
saltgen: I think you are having this concern cause of the exception in the validator for `updated,` but we will remove that in future.
Both the engine and events consumer has two main requirements `is_ros` and `cloud_provider`, hence apart from the exception I don't see much difference w.r.t create and update from validator pov.

ros/lib/utils.py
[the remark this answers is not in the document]
saltgen: Yes, cause I believe puptoo still sends is_ros: 'true', so just to support that

ros/processor/inventory_events_consumer.py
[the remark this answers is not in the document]
kgaikwad: Account record creation is happening above this check [here](https://github.com/RedHatInsights/ros-backend/pull/316/files/4f5779b8a1ee76816cc6f2e221ac52c651a05a15#diff-3ec10f47264a06a134d0bf1ce745fb26ad086de0af6da0cf6e83c00dcf2c4193R130).

kgaikwad:
@saltgen - could you please perform rebase for this PR? 
As RHIROS-1112 changes already handled in a separate PR, no longer needed to include that. 
You might need to update commit message again. :(

PR #333, opened 2023-05-31
Fixes #RHIROS-1155 - custom readonly recommendation class

kgaikwad:
## PR Title :boom:

Suggested formats: 

2. RHIROS-XXX Title 

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:


ros/api/common/recommendation_model.py
```diff
@@ -0,0 +1,74 @@
...
+        summaries = [
+            ROSSUMMARY[state] for substates in states.values()
+            for state in substates
+            if ROSSUMMARY.get(state) is not None
+        ]
+        if rule_hit_key == 'INSTANCE_IDLE':
+            summaries = None
+
+        if summaries is None:
+            return None
```
upadhyeammit, line 72:
```suggestion
        if rule_hit_key == 'INSTANCE_IDLE':
            return None
        else:
            summaries = [
            ROSSUMMARY[state] for substates in states.values()
            for state in substates
            if ROSSUMMARY.get(state) is not None
        ]

```

ros/api/common/recommendation_model.py
```diff
@@ -0,0 +1,74 @@
...
+)
+
+
+class Recommendation:
+    """Custom readonly recommendation class to generate list."""
+    def __init__(self, rule_data, rule_hit, system, psi_enabled):
+        rule_hit_key = rule_hit.get("key")
+
+        rule_dict = rule_data.__dict__
+        if system.cloud_provider is None:
```
upadhyeammit, line 31: There should be no recommendation if cloud provider is None? and I think we should rely on the recently added logic which is about if system is allowed in ros?
https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py#L270

upadhyeammit (changes requested):
I am not convinced with the file and directory naming(`ros/api/common/recommendation_model.py`) however at the same time I am unsure what can be correct dir structure and filename?

Thanks a lot for separating the code in small methods!

ros/api/common/recommendation_model.py
kgaikwad: This was added as a temporary fix as part of [PR](https://github.com/RedHatInsights/ros-backend/pull/219#pullrequestreview-1028527833) Original issue was on Inventory side where it stopped sending cloud_provider value.
As mentioned by you, yes, we can remove it now!

kgaikwad:
> I am not convinced with the file and directory naming(`ros/api/common/recommendation_model.py`) however at the same time I am unsure what can be correct dir structure and filename?

I tried to find guidelines but there is no any fix standard provided by python. I am happy to move under any other directory, please let me know if you have any suggestions on file naming and directory.

ros/api/common/recommendation_model.py
r14chandra, line 1: IMO this file should reside under `api` dir and not under`../api/common` as `common` seems like it contains files/modules that would be used in other places but here this module would only be used in recommendation.py file. 

Also, we already have a model for recommendation so the `recommendation_model` file name seems redundant. We can rename it as `recommendation_query.py`.

r14chandra (changes requested):

ros/api/common/recommendation_model.py
saltgen, line 1: @kgaikwad IMO, we could do something like so,

```
ros_backend/
├── modules/
│   ├── recommendations.py
│   └── ...
└── api/
    ├── recommendations.py
    └── ...
```
This way we can move other classes in different files and move them to `modules` directory.
Additionally the file names make sense in their own context i.e. directory and hence imports look cleaner as well.

`from api import recommendations`

`from utils import recommendations`

saltgen (changes requested):

ros/api/common/recommendation_model.py
kgaikwad: >  we already have a model for recommendation so the recommendation_model file name seems redundant.

@r14chandra - IIRC, we don't have recommendation model.  


> this file should reside under api dir and not under../api/common

I can keep it to same directory i.e under `api`. But let me know your thoughts on what @saltgen suggested [here](https://github.com/RedHatInsights/ros-backend/pull/333#discussion_r1221106443). 
In that way it would keep api files and module files separate.

kgaikwad: Keeping `modules` folder under ros > api directory.

r14chandra (approved):
Looks good!

PR #344, opened 2023-07-24
feat: Add ability to connect to DB with ssl

SteveHNH:
This is a requirement for some environments.

- Added a check for the clowder configured RDS CA
- Tack the SSL mode onto the DB URI if it exists

## PR Title :boom:

I don't have a JIRA for this one. Just noticed this as an issue in one of our environments that requires SSL so I took the "patches welcome" approach

Some environments require SSL connectivity to the RDS databases. Clowder can provide this cert so we should look for it if it exists

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [x] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [x] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

This is just a suggestion for how to do this that should get SSL support into the code. If there are other patterns that the ROS team would like to use, that's fine by me.

### DIFF
--- a/ros/lib/config.py
+++ b/ros/lib/config.py
@@ -45,6 +45,8 @@ def build_endpoint_url(ep):
 
 LOG = logging.getLogger(__name__)
 CLOWDER_ENABLED = True if os.getenv("CLOWDER_ENABLED", default="False").lower() in ["true", "t", "yes", "y"] else False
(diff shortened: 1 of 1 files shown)


kgaikwad:
Thank you @SteveHNH.

PR #346, opened 2023-07-25
Fixes #RHIROS-1266 - Add groups support through RBAC

upadhyeammit:

The RBAC now returns groups to which user has access to. The System and dependent objects should be gated as per the group details.

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [x] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/ros/api/common/add_group_filter.py
+++ b/ros/api/common/add_group_filter.py
@@ -1,0 +1,10 @@
+from ros.lib.models import System
+
+
+def include_group_filter(request, query):
+    if hasattr(request, "host_groups"):
+        if len(request.host_groups) > 1 or (len(request.host_groups) == 0 and request.host_groups[0] is not None):
+            query = query.filter(System.groups[0]['id'].astext.in_(request.host_groups) | (System.groups == '[]'))
+        else:
+            query = query.filter((System.groups == '[]'))
+    return query
--- a/ros/api/common/add_group_filter.py
+++ b/ros/api/common/add_group_filter.py
@@ -1,7 +1,8 @@
 from ros.lib.models import System
+from flask_restful import request
 
 
-def include_group_filter(request, query):
+def group_filtered_query(query):
     if hasattr(request, "host_groups"):
         if len(request.host_groups) > 1 or (len(request.host_groups) == 0 and request.host_groups[0] is not None):
             query = query.filter(System.groups[0]['id'].astext.in_(request.host_groups) | (System.groups == '[]'))
--- a/ros/api/common/add_group_filter.py
+++ b/ros/api/common/add_group_filter.py
@@ -4,8 +4,9 @@ from flask_restful import request
 
 def group_filtered_query(query):
     if hasattr(request, "host_groups"):
-        if len(request.host_groups) > 1 or (len(request.host_groups) == 0 and request.host_groups[0] is not None):
-            query = query.filter(System.groups[0]['id'].astext.in_(request.host_groups) | (System.groups == '[]'))
+        host_groups = [gid for gid in request.host_groups if gid is not None]
+        if len(host_groups) > 1:
+            query = query.filter(System.groups[0]['id'].astext.in_(host_groups) | (System.groups == '[]'))
         else:
             query = query.filter((System.groups == '[]'))
     return query
--- a/ros/lib/rbac_interface.py
+++ b/ros/lib/rbac_interface.py
@@ -5,6 +5,7 @@ from .config import RBAC_SVC_URL, ENABLE_RBAC, TLS_CA_PATH
 import requests
 import json
 from ros.lib.config import get_logger
+from flask import request
 
 RBAC_SVC_ENDPOINT = "/api/rbac/v1/access/?application=%s"
 AUTH_HEADER_NAME = "X-RH-IDENTITY"
@@ -75,11 +76,6 @@ def query_rbac(application, auth_key, logger):
     return rbac_result
 
 
-def get_perms(result):
-    perms = [perm["permission"] for perm in result["data"]]
-    return perms
-
-
 def ensure_has_permission(**kwargs):
     """
     Ensure permission exists. kwargs needs to contain:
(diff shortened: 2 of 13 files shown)


kgaikwad (changes requested):

ros/api/common/add_group_filter.py
```diff
@@ -0,0 +1,10 @@
+from ros.lib.models import System
+
+
+def include_group_filter(request, query):
+    if hasattr(request, "host_groups"):
+        if len(request.host_groups) > 1 or (len(request.host_groups) == 0 and request.host_groups[0] is not None):
```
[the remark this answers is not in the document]
upadhyeammit: As per the doc RBAC returns json's null when there are no groups for the user. So if user is not having any group null would be there in the group.id values. In that case if we parse the json it becomes None and finally list becomes [None]

upadhyeammit:
Just for the curiosity, when we return list of the system we also depend on if it has the performance profile associated. Is it necessary to only return the systems which are having performance profile when we do get systems api call?

kgaikwad (changes requested):

ros/lib/rbac_interface.py
```diff
@@ -118,3 +120,79 @@ def _is_mgmt_url(path):
     Small helper to test if URL is for management API.
     """
     return path.startswith("/mgmt/")
+
+
+def set_host_groups(rbac_response):
+    """
+    We now also have to store the host group information we get from inventory.
```
PreetiW, line 127: @upadhyeammit based on my understanding the text should be `store the user group information we get from RBAC?` correct me if I am wrong 🙌🏻

ros/lib/rbac_interface.py
```diff
@@ -118,3 +120,79 @@ def _is_mgmt_url(path):
...
+    This comes in the resource definition within the RBAC response:
+
+    {
+      "resourceDefinitions": [
+        {
+          "attributeFilter": {
+            "key": "inventory.groups",
```
PreetiW, line 134: @upadhyeammit [based on this doc](https://docs.google.com/document/d/1O87MHvKlPEoH_DHwZw4htvwYlkGMiS-LwLQ_qKsZ5J0/edit#heading=h.1z0ggwl751hw) the key should be `group.id` ? or am I missing something here?

ros/lib/rbac_interface.py
```diff
@@ -118,3 +120,79 @@ def _is_mgmt_url(path):
...
+            ],
+            "operation": "in"
+          }
+        }
+      ],
+      "permission": "inventory:hosts:read"
+    }
+
+    If the permission is 'inventory:hosts:read', we find one of the resource
+    definitions that has an attribute filter key of 'inventory.groups', and
```
PreetiW, line 147: same as [above comment](https://github.com/RedHatInsights/ros-backend/pull/346/files#r1277380865)

ros/lib/rbac_interface.py
```diff
@@ -118,3 +120,79 @@ def _is_mgmt_url(path):
...
+            # versions say it's a string with a JSON-encoded list.  Let's try
+            # to cope with the latter by converting it into the former.
+            if isinstance(value, str) and value[0] == '[' and value[-1] == ']':
+                value = json.loads(value)
+            if not isinstance(value, list):
+                continue
+            # Finally, we have the right key: set its value (a list) as
+            # the 'host_group_attr' property
+            # Maybe it would be good to set values as json here instead of list
+            setattr(request, host_group_attr, value)
```
PreetiW, line 195: @upadhyeammit I am just curious to know as I have limited knowledge of python that are we setting a att for any request we receive on the backend side for groups? and then we will access the group info in the query?

tests/test_api_endpoints.py
```diff
@@ -440,3 +411,73 @@ def test_openapi_endpoint():
...
+        system_with_test_group,
+        db_create_performance_profile,
+        create_performance_profiles,
+        mocker):
+    with app.test_client() as client:
+        mock_enable_rbac(mocker)
+        mock_rbac(get_rbac_mock_file("mock_rbac_returns_no_groups.json"), mocker)
+        response = client.get('/api/ros/v1/systems', headers={"x-rh-identity": auth_token})
+        assert response.status_code == 200
+        assert response.json["meta"]["count"] == 3
```
PreetiW, line 483: @upadhyeammit do you think we should add comments above each test case to understand what scenario it is testing? like what will be expected output to a particular scenario?

PreetiW (commented):
@upadhyeammit Added few inline comments, overall the PR looks great. Once the comments are resolved we can go ahead and merge the PR. Thank you for much for your work on this 🙌🏻

ros/lib/rbac_interface.py
upadhyeammit: Yeah the next line says that exactly:
~~~
We now also have to store the host group information we get from inventory.
 This comes in the resource definition within the RBAC response:
~~~

ros/lib/rbac_interface.py
```diff
@@ -118,3 +120,79 @@ def _is_mgmt_url(path):
...
+    We now also have to store the host group information we get from inventory.
+    This comes in the resource definition within the RBAC response:
+
+    {
+      "resourceDefinitions": [
+        {
+          "attributeFilter": {
+            "key": "inventory.groups",
+            "value": [
+              "group 1",
```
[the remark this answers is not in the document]
upadhyeammit: Its comment section, so I am keeping the unchanged!

ros/lib/rbac_interface.py
upadhyeammit: Yea you are correct. We are setting an extra attribute on request object! initially my thoughts were somehow returning or exposing this group list using var or method for each request, however patching a new attribute on existing object felt better in terms of understanding, 
1. Its per request group list
2. Carry it cleanly with each request object

PreetiW: @upadhyeammit got it thanks for sharing the details 🙌🏻

PreetiW (approved):
Everything looks good now, thank you so much @upadhyeammit for adding test description it is very helpul. LGTM 💯

upadhyeammit:
Thanks @PreetiW and @kgaikwad for quick reviews! I am merging this one!

PR #354, opened 2023-08-24
Fixes #RHIROS-1269 - feature flags for backend

upadhyeammit:
## PR Title :boom:

Suggested formats: 

2. RHIROS-XXX Title 

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/clowdapp.yaml
+++ b/clowdapp.yaml
@@ -10,6 +10,7 @@ objects:
     name: ros-backend
   spec:
     envName: ${ENV_NAME}
+    featureFlags: true
     dependencies:
     - host-inventory
     - rbac
@@ -63,6 +64,17 @@ objects:
             value: ${DB_POOL_SIZE}
           - name: DB_MAX_OVERFLOW
             value: ${DB_MAX_OVERFLOW}
+          - name: UNLEASH_URL
+            value: ${UNLEASH_URL}
+          - name: UNLEASH_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: ${UNLEASH_SECRET_NAME}
+                key: CLIENT_ACCESS_TOKEN
+                optional: true
+          - name: BYPASS_UNLEASH
+            value: ${BYPASS_UNLEASH}
+
     - name: inventory-events-processor
       replicas: ${{INVENTORY_PROCESSOR_REPLICA_COUNT}}
       podSpec:
@@ -103,6 +115,17 @@ objects:
             value: ${DB_POOL_SIZE}
           - name: DB_MAX_OVERFLOW
             value: ${DB_MAX_OVERFLOW}
+          - name: UNLEASH_URL
+            value: ${UNLEASH_URL}
+          - name: UNLEASH_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: ${UNLEASH_SECRET_NAME}
+                key: CLIENT_ACCESS_TOKEN
+                optional: true
+          - name: BYPASS_UNLEASH
+            value: ${BYPASS_UNLEASH}
+
     - name: engine-result-processor
       replicas: ${{ENGINE_PROCESSOR_REPLICA_COUNT}}
       podSpec:
@@ -143,6 +166,17 @@ objects:
             value: ${DB_POOL_SIZE}
           - name: DB_MAX_OVERFLOW
             value: ${DB_MAX_OVERFLOW}
+          - name: UNLEASH_URL
+            value: ${UNLEASH_URL}
+          - name: UNLEASH_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: ${UNLEASH_SECRET_NAME}
+                key: CLIENT_ACCESS_TOKEN
+                optional: true
+          - name: BYPASS_UNLEASH
+            value: ${BYPASS_UNLEASH}
+
     - name: garbage-collector-processor
       replicas: ${{GARBAGE_COLLECTOR_REPLICA_COUNT}}
       podSpec:
@@ -187,6 +221,17 @@ objects:
             value: ${DB_POOL_SIZE}
           - name: DB_MAX_OVERFLOW
             value: ${DB_MAX_OVERFLOW}
+          - name: UNLEASH_URL
+            value: ${UNLEASH_URL}
+          - name: UNLEASH_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: ${UNLEASH_SECRET_NAME}
+                key: CLIENT_ACCESS_TOKEN
+                optional: true
+          - name: BYPASS_UNLEASH
+            value: ${BYPASS_UNLEASH}
+
     database:
       name: ros
       version: 13
@@ -305,3 +350,13 @@ parameters:
   value: "20"
 - name: DB_MAX_OVERFLOW
   value: "20"
+
+# Feature flags
+- description: Unleash secret name
+  name: UNLEASH_SECRET_NAME
+  value: bypass
+- description: Unleash API url
+  name: UNLEASH_URL
+- description: disable Unleash (feature flags), defaulting to fallback values
+  name: BYPASS_UNLEASH
+  value: 'false'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ watchtower = ">=1.0.0"
 boto3 = ">=1.16.13"
 botocore = ">=1.19.13"
 flask-caching = "*"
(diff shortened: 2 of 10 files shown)


upadhyeammit:
I have few questions,
1. We are using fallback value for unleash flag, this means that if unleash is enabled and for some reason unleash service does not response then we fallback to some value. Now this value is True. This means if unleash is enabled anyhow we are going to consider its enabled even when the unleash service itself is not able to respond. This is what I can see inventory is following as well. Is this fine?
2. Inventory has bypassed the unleash in the test env, however for us at least I want to execute the groups related unit tests and so I have not bypassed it. This is why we are mocking the unleash enablement. Is this fine?
3. Locally by default unleash has been bypassed. Is this fine?
4. If unleash is available and for some reasons rbac ended up not giving us the group details, in that case the current scenario is we show the systems which are not in any group, is this acceptable?

PreetiW:
@upadhyeammit do we know why the PR checks are failing or do we want QE to look into it?

upadhyeammit:
> @upadhyeammit do we know why the PR checks are failing or do we want QE to look into it?

Its failing for cache dir permissions, I am having a look!

upadhyeammit:
> > @upadhyeammit do we know why the PR checks are failing or do we want QE to look into it?
> 
> Its failing for cache dir permissions, I am having a look!

I removed it, its not mandatory however I was expecting that it would work

https://docs.getunleash.io/Flask-Unleash/

upadhyeammit:
> > > @upadhyeammit do we know why the PR checks are failing or do we want QE to look into it?
> > 
> > 
> > Its failing for cache dir permissions, I am having a look!
> 
> I removed it, its not mandatory however I was expecting that it would work
> 
> https://docs.getunleash.io/Flask-Unleash/

It seems this doesn't work either, now its creating it under /.cache, checking with appsre

PreetiW:
@upadhyeammit can you please rebase this branch?

ros/api/common/add_group_filter.py
```diff
@@ -1,12 +1,24 @@
...
+
+logger = get_logger(__name__)
 
 
 def group_filtered_query(query):
-    if hasattr(request, "host_groups"):
-        host_groups = [gid for gid in request.host_groups if gid is not None]
-        if len(host_groups) >= 1:
-            query = query.filter(System.groups[0]['id'].astext.in_(host_groups) | (System.groups == '[]'))
+    if get_flag_value(FLAG_INVENTORY_GROUPS):
```
PreetiW, line 10: @upadhyeammit according to me the feature flag checks should be at the API level as well and for RBAC if feature flag is enabled it should follow the condition based on RBAC response else we should skip adding any filter to the query is what I feel from the functionality point of view so that it will return all the systems and work like before. Let me know your thoughts!

PreetiW: What I mean by API level is that if feature flag is disabled that group info should not come from the Systems API response and filtering and sorting on `group_name` should not work.

pyproject.toml
```diff
@@ -27,6 +27,7 @@ watchtower = ">=1.0.0"
 boto3 = ">=1.16.13"
 botocore = ">=1.19.13"
 flask-caching = "*"
+flask-unleash = "^2.0.0"
```
PreetiW, line 30: @upadhyeammit is inventory also using the same package? I am asking as it looks like t[he package is not maintained recently.](https://github.com/Unleash/Flask-Unleash)

clowdapp.yaml
```diff
@@ -305,3 +350,13 @@ parameters:
...
+
+# Feature flags
+- description: Unleash secret name
+  name: UNLEASH_SECRET_NAME
+  value: bypass
+- description: Unleash API url
+  name: UNLEASH_URL
+- description: disable Unleash (feature flags), defaulting to fallback values
+  name: BYPASS_UNLEASH
+  value: 'false'
```
PreetiW, line 362: @upadhyeammit can you please explain a bit about the changes we did here for my understanding?

PreetiW (changes requested):
Thank you for creating PR @upadhyeammit, I have left a few comments. Let me know if any questions, overall PR looks good 💯

PreetiW:
@upadhyeammit here are my thoughts 💯 

> I have few questions,
> 
> 1. We are using fallback value for unleash flag, this means that if unleash is enabled and for some reason unleash service does not response then we fallback to some value. Now this value is True. This means if unleash is enabled anyhow we are going to consider its enabled even when the unleash service itself is not able to respond. This is what I can see inventory is following as well. Is this fine?  **According to me Yes 👍🏻** 
> 2. Inventory has bypassed the unleash in the test env, however for us at least I want to execute the groups related unit tests and so I have not bypassed it. This is why we are mocking the unleash enablement. Is this fine? **+1 with this**
> 3. Locally by default unleash has been bypassed. Is this fine? **+1 with this, I don't see any need of using unleash locally**
> 4. If unleash is available and for some reasons rbac ended up not giving us the group details, in that case the current scenario is we show the systems which are not in any group, is this acceptable? **I have added comments regarding how I think we need to use unleash feature flag, have a look and lets continue the discussion in that thread.**

ros/api/common/add_group_filter.py
upadhyeammit: As per new changes if unleash is not enabled then we return empty groups array instead of completely getting rid of groups field from the response. This is to keep in sync with openapi spec

ros/api/v1/hosts.py
```diff
@@ -257,7 +260,7 @@ def build_sort_expression(self, order_how, order_method):
             return (sort_order(PerformanceProfile.report_date),
                     asc(PerformanceProfile.system_id))
 
-        if order_method == 'group_name':
+        if order_method == 'group_name' and get_flag_value(FLAG_INVENTORY_GROUPS):
```
PreetiW, line 263: @upadhyeammit shall we revert this change similar to above group_name filter?

ros/api/v1/hosts.py
```diff
@@ -319,6 +322,8 @@ def get(self, host_id):
             record['os'] = system.deserialize_host_os_data
             record['number_of_recommendations'] = profile.number_of_recommendations
             record['psi_enabled'] = profile.psi_enabled
+            if not get_flag_value(FLAG_INVENTORY_GROUPS):
+                record['groups'] = []
```
PreetiW, line 326: @upadhyeammit can you please share why we added this code? If I am not wrong this is for performance profile? If yes I am not sure if we need group info there.

tests/test_api_endpoints.py
```diff
@@ -115,32 +117,40 @@ def test_system_groups(auth_token, db_setup, db_create_account, db_create_system
...
+        "id": "12345678-fe1b-4191-8408-cbadbd47f7a3",
         "name": "ros-test-3"
     }]
     db.session.commit()
 
     with app.test_client() as client:
+        mock_enable_rbac(mocker)
+        mock_rbac(get_rbac_mock_file("mock_rbac_returns_groups_including_example_group.json"), mocker)
+
+        mock_unleash_hbi_flag_enabled(mocker)
```
PreetiW, line 129: @upadhyeammit is it possible to add a test case when unleash flag is false and check for default values for groups in that case?

PreetiW (commented):
@upadhyeammit Thank you for making the changes, just added minor comments on the PR. Rest is looking good 💯

ros/api/v1/hosts.py
upadhyeammit: yeah!

PreetiW: @upadhyeammit I agree we are not assigning anything to the order method, I just felt that checking should be done after feature flag checking. But if you are okay with this, I am okay as well 👍🏻

ros/api/v1/hosts.py
upadhyeammit: I think then maybe we don't even need the Groups column for Executive report?

PreetiW: @upadhyeammit I don't see any use case ATM where it will be useful in Executive report so I feel we dont need it there. But let me know if you can see any usecase.

upadhyeammit: I removed it from the pprofile.

PreetiW (approved):
Thank you for changes @upadhyeammit , LGTM 🚀

PR #379, opened 2023-10-04
Pin werkzeug to <3.0.0

r14chandra:


upadhyeammit (approved):
Request to update commit msg something like: Pin werkzeug to <3.0.0 .
and change makes sense, thanks!

PR #380, opened 2023-10-04
Pin flask_caching to <=2.0.2

r14chandra:
`AttributeError: 'RedisCache' object has no attribute '_client'`
The above is caused by `flask_caching` [v1.11.1](https://github.com/pallets-eco/flask-caching/pull/372) and is fixed in v2.0.0.
To upgrade it to the latest one we need to pin `flask` <v3.0.0 as its latest version (2.0.2) depends on Flask (<3). 
And in turn, we would need to pin `werkzeug` <v3.0.0

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1695,4 +1695,4 @@ testing = ["big-O", "jaraco.functools", "jaraco.itertools", "more-itertools", "p
 [metadata]
 lock-version = "2.0"
 python-versions = "~=3.8"
-content-hash = "7caa6940e9f302e785eea832ef0567a7732279bdacdc6abb131312dd1909f518"
+content-hash = "574e63b42373df90b46de98e1011e79cf1e0f3ce390ffbaa4e7241b2a7827f64"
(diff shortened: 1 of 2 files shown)


upadhyeammit:
@r14chandra my thoughts on this,

TL;DR: I think these extra pipelines are there as indicator for us to know if we can update to newer version of dependencies without breaking anything and without major change. If we pin the versions in that then it would not give us that benefit?

We have the extra pipelines to exactly do what you are doing however it is to help in bumping versions in the pyproject file, I mean in production.

If we go by this road, I mean pinning the versions for extra gh pipes then we would also need to keep updating them once we get newer versions of dependencies, and if we miss on this then it would actually defeat the benefit of extra gh action? I mean, now as you have pinned the versions we don't know if its breaking for next upcoming version as well?

If we dont go by this road, no pinning at all. We would always know its failing for the latest ones and that means,
1. Our app is not outright ready for consuming the latest packages
2. We would need to see the version dependencies and pin the stuff if we want to get it working for latest ones

upadhyeammit (approved):

PR #382, opened 2023-10-09
Fixes #RHIROS-1176, #RHIROS-1178 - suggested_instance_types/<instance_type> api endpoint

upadhyeammit:
The suggested_instance_types api endpoint accepting instance_type with filtering defined on display_name, state and OS

## PR Title :boom:

Suggested formats: 

2. RHIROS-XXX Title 

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/ros/api/v1/suggested_instance_types.py
+++ b/ros/api/v1/suggested_instance_types.py
@@ -22,15 +22,6 @@ from ros.api.common.pagination import (
 from ros.api.common.instance_types_helper import instance_types_desc_dict
 from ros.api.common.utils import sorting_order
 
-SYSTEM_DETAILS_COLUMNS = [
-    'fqdn',
-    'display_name',
-    'instance_type',
-    'os',
-    'performance_utilization',
-    'state',
-    'last_reported'
-]
 LOG = logging.getLogger(__name__)
 
 
--- a/ros/api/v1/suggested_instance_types.py
+++ b/ros/api/v1/suggested_instance_types.py
@@ -1,15 +1,17 @@
+import logging
 from flask import request
 from sqlalchemy import func
 from flask_restful import Resource, fields, marshal_with, abort
 from ros.api.common.add_group_filter import group_filtered_query
 from ros.lib.utils import (
     system_ids_by_org_id,
-    org_id_from_identity_header,
+    org_id_from_identity_header, sort_io_dict,
 )
 
 from ros.lib.models import (
     db,
     PerformanceProfile,
+    System
 )
 
 from ros.api.common.pagination import (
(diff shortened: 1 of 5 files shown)


ros/api/v1/suggested_instance_types.py
```diff
@@ -96,3 +109,104 @@ def build_sort_expression(self, order_how, order_method):
...
+    @marshal_with(output)
+    def get(self, instance_type):
+        limit = limit_value()
+        offset = offset_value()
+        org_id = org_id_from_identity_header(request)
+        system_query = group_filtered_query(system_ids_by_org_id(org_id))
+        query = db.session.query(
+            System.fqdn,
+            System.display_name,
+            func.concat(System.operating_system["name"].astext, ' ', System.operating_system['major'], '.',
```
r14chandra, line 166: We have [this](https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/models.py#L90) function already to deserialize OS data. We can use that.

ros/api/v1/suggested_instance_types.py
```diff
@@ -96,3 +109,104 @@ def build_sort_expression(self, order_how, order_method):
...
+
+    @staticmethod
+    def build_instance_details_filters():
+        filters = []
+        if filter_display_name := request.args.get('display_name'):
+            filters.append(System.display_name.ilike(f'%{filter_display_name}'))
+        if filter_state := request.args.get('state'):
+            filters.append(System.state.ilike(f'%{filter_state}'))
+        if filter_os := request.args.get('os'):
+            filters.append(func.concat(System.operating_system["name"].astext, ' ',
```
r14chandra, line 209: Same here. We can use the custom deserialize method.

r14chandra (changes requested):

ros/api/v1/suggested_instance_types.py
upadhyeammit: While trying to use the `deserialize_host_os_data` function I faced below challenges,
1. Its the property of System's model
2. If I want to use it then in query I would need to select all columns of System's model(actually call whole model) even though I don't need all of them

So I have below thoughts,
a. We retrieve all columns from tables and use the function, which I think we should not do?
b. Rewrite the `deserialize_host_os_data` but not as property, this means we would also change existing code which depends on this method.

Let me know your thoughts

r14chandra: The second one is better. But I don't know much changes are required in the existing code. We can target it in separate PR or in the same. What do you think?

upadhyeammit: As per the discussion we would like to take route b, which would need more code change. Considering this I would have separate pr for that refactoring

r14chandra (approved):

r14chandra:
@yash2189 Could test the PR before we merge?

yash2189:
Tests are successful for this PR and observations have been added over respective Jira card.

upadhyeammit:
Thanks a lot for reviews and testing! Merging this now!

PR #390, opened 2023-11-01
RHIROS-715 add contextual filters on log records

r14chandra:


ros/lib/cw_logging.py
```diff
@@ -58,4 +60,32 @@ def commence_cw_log_streaming(stream_name):
         logger.info(f"{module_prefix} - Streaming in progress - Log group: {AWS_LOG_GROUP}")
         watchtower_handler.setLevel(logging.INFO)
         watchtower_handler.setFormatter(logging.Formatter(fmt=CW_LOGGING_FORMAT))
+        watchtower_handler.setFormatter(LogstashFormatterV1())
+        watchtower_handler.addFilter(ContextualFilter())
         root_logger.addHandler(watchtower_handler)
+
+
+class ContextualFilter(logging.Filter):
```
upadhyeammit, line 68: I tried to go to definition of addFilter and follow along to python's code. However still I am unable to conclude if this requires class or not?
Is it possible to just have the filter method and not the ContextualFilter?

r14chandra: I have used one of the ways suggested in [this](https://docs.python.org/3/howto/logging-cookbook.html#using-filters-to-impart-contextual-information) doc. Moreover, HBI, Puptoo, and Yuptoo are also following the same.

kgaikwad:
@r14chandra - could you please rebase the pull-request?

upadhyeammit (approved):

PR #404, opened 2023-11-30
Update labels as required for RHTAP

upadhyeammit:

If we don't have LABELS defined then by default image creation inherits it from parent image which is UBI. This is not acceptable by RHTAP. So defining the LABELS for ROS image.

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,3 +22,21 @@ COPY seed.d seed.d
 RUN python -m venv $VIRTUAL_ENV && \
     poetry update pip setuptools wheel && \
     poetry install
+
+LABEL description="The Red Hat Insights resource optimization service \
+enables RHEL customers to assess and monitor their public cloud usage \
+and optimization. The service exposes workload metrics for CPU, memory, \
+and disk-usage and compares them to resource limits recommended by the \
+public cloud provider. Currently ROS only provides suggestions for AWS \
+RHEL instances. To enable ROS, a customer needs to perform a few \
+prerequisite steps on targeted systems via Ansible playbook."
+LABEL io.k8s.description="The Red Hat Insights resource optimization service \
+enables RHEL customers to assess and monitor their public cloud usage \
+and optimization. The service exposes workload metrics for CPU, memory, \
+and disk-usage and compares them to resource limits recommended by the \
+public cloud provider. Currently ROS only provides suggestions for AWS \
+RHEL instances. To enable ROS, a customer needs to perform a few \
+prerequisite steps on targeted systems via Ansible playbook."
+LABEL io.k8s.display-name="ros-backend"
+LABEL io.openshift.tags="ros-backend"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -22,3 +22,21 @@ COPY seed.d seed.d
...
+LABEL io.k8s.description="The Red Hat Insights resource optimization service \
+enables RHEL customers to assess and monitor their public cloud usage \
+and optimization. The service exposes workload metrics for CPU, memory, \
+and disk-usage and compares them to resource limits recommended by the \
+public cloud provider. Currently ROS only provides suggestions for AWS \
+RHEL instances. To enable ROS, a customer needs to perform a few \
+prerequisite steps on targeted systems via Ansible playbook."
+LABEL io.k8s.display-name="ros-backend"
+LABEL io.openshift.tags="ros-backend"
+LABEL summary="Image of Resource Optimization Service backend"
```
kgaikwad, line 42: In older version of docker, to prevent extra layers it was recommended to combine into single `LABEL` instruction. With newer versions, it is not necessary but still supported. So, do you think these instructions can be combined into single instruction?

Dockerfile
```diff
@@ -22,3 +22,21 @@ COPY seed.d seed.d
...
+public cloud provider. Currently ROS only provides suggestions for AWS \
+RHEL instances. To enable ROS, a customer needs to perform a few \
+prerequisite steps on targeted systems via Ansible playbook."
+LABEL io.k8s.description="The Red Hat Insights resource optimization service \
+enables RHEL customers to assess and monitor their public cloud usage \
+and optimization. The service exposes workload metrics for CPU, memory, \
+and disk-usage and compares them to resource limits recommended by the \
+public cloud provider. Currently ROS only provides suggestions for AWS \
+RHEL instances. To enable ROS, a customer needs to perform a few \
+prerequisite steps on targeted systems via Ansible playbook."
```
kgaikwad, line 39: It seems both `description` label values are same, could you declare variable and use at both places? 
Also, as these labels required just for RHTAP and IIUC this integration for our internal use, in my opinion we can keep short meaningful description for it. Thoughts?

kgaikwad (changes requested):
Hello @upadhyeammit, thank you for a change. Added few suggestions inline.

Dockerfile
upadhyeammit: Yeah we can have short description, however I just felt it should be in sync with description of our repository.

upadhyeammit: I am merging this now! if require we would revisit this

PR #410, opened 2023-12-21
Fixes #RHINENG-5908 - handle updated platform_metadata dict

upadhyeammit:

The platform_metadata use to be empty if there is nothing in it, however now when its empty it at least has "bs64_identity". This breaks ROS use cases as we use to depend on "platform_metadata" being None!

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [x] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/ros/lib/utils.py
+++ b/ros/lib/utils.py
@@ -350,8 +350,18 @@ def system_allowed_in_ros(msg, reporter):
         # https://consoledot.pages.redhat.com/docs/dev/services/inventory.html#_updated_event
         if (
                 msg.get('type') == 'updated'
-                and msg.get('platform_metadata') is None
+                and is_platform_metadata_empty(msg)
         ):
             return is_valid_cloud_provider(cloud_provider)
         is_ros = msg["platform_metadata"].get("is_ros")
     return validate_ros_payload(is_ros, cloud_provider, operating_system)
+
+
+def is_platform_metadata_empty(msg):
+    if msg.get("platform_metadata") is None or \
+            (isinstance(msg.get("platform_metadata"), dict) and
+             len(msg.get("platform_metadata")) < 1 and
+             "b64_identity" in msg.get("platform_metadata")):
+        return True
+
+    return False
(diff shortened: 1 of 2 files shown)


upadhyeammit:
With this kind of breaking change I do feel we should have centralized way to define on what all params we work or on depend on and actually have better validations on it.

ros/lib/utils.py
```diff
@@ -350,8 +350,18 @@ def system_allowed_in_ros(msg, reporter):
...
                 msg.get('type') == 'updated'
-                and msg.get('platform_metadata') is None
+                and is_platform_metadata_empty(msg)
         ):
             return is_valid_cloud_provider(cloud_provider)
         is_ros = msg["platform_metadata"].get("is_ros")
     return validate_ros_payload(is_ros, cloud_provider, operating_system)
+
+
+def is_platform_metadata_empty(msg):
```
kgaikwad, line 360:
```suggestion
def is_platform_metadata_check_pass(msg):
```

Could you rename the method something similar ^^ as no longer empty check?

kgaikwad (changes requested):
Added small inline suggestion otherwise I think PR looks good.

ros/lib/utils.py
```diff
@@ -350,8 +350,18 @@ def system_allowed_in_ros(msg, reporter):
...
+    if msg.get("platform_metadata") is None or \
+            (isinstance(msg.get("platform_metadata"), dict) and
+             len(msg.get("platform_metadata")) < 1 and
```
PreetiW, line 363: @upadhyeammit we will need to update this condition as `<=1` as we are getting `b64_identity` in the msg else it will not work. Let me know if I got it incorrect 🙌🏻

PreetiW (changes requested):
@upadhyeammit thank you for making the changes, I have tested the PR and based on it left a comment 👍🏻 
As discussed once you can confirm the `platform_metadata` structure from the inventory team is same as what we have assumed in this PR, post that I feel it will be good to merge this 💯

ros/lib/utils.py
kgaikwad: Good catch!

kgaikwad (approved):

PR #433, opened 2024-05-22
Enable Platform-Security GitHub Workflow

casey-williams-rh:
## Overview
Enabling the `Platform Security GitHub Workflow` - This workflow provides Red Hat ConsoleDot Teams with a way to scan the containers they create in a convenient, automated, and reliable manner within their GitHub repository. The Platform Security Github Workflow lets teams get security feedback as they open Pull Requests.

Additionally, the `FROM` statement in the `Dockerfile` has been changes to `registry.access.redhat.com` to ensure that the security scanner and pull the Base Image without needing to authenticate with the RH Registry. 

REF: https://github.com/RedHatInsights/platform-security-gh-workflow

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.redhat.io/ubi8/ubi-minimal
+FROM registry.access.redhat.com/ubi8/ubi-minimal
 
 ARG PYTHON_PIP_VERSION=23.3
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.redhat.io/ubi8/ubi-minimal
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
upadhyeammit, line 1: Is it required to use `registry.access.redhat.com`?

casey-williams-rh: In order for the GH Workflow to pull the base image to build the app image and scan it. We need to use `registry.access.redhat.com` as it is the open image repo for Red Hat. `registry.redhat.io` requires us to provide credentials to pull the image to build and scan. It is the same image, just a different path to pull it.

upadhyeammit (approved):

PR #466, opened 2024-08-28
Fixes #RHINENG-12168 - add created_at field to rh_accounts

kgaikwad:
## PR Title :boom:

Suggested formats: 

2. RHIROS-XXX Title 

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:


ros/lib/models.py
```diff
@@ -132,6 +132,8 @@ class RhAccount(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     account = db.Column(db.Text, nullable=True)
     org_id = db.Column(db.Text)
+    created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow)
+    updated_at = db.Column(db.DateTime, default=datetime.datetime.utcnow)
```
upadhyeammit, line 136: The pull  request and Jira says its about just adding the fields. So is there another pull request for the function and trigger to update updated_at?

kgaikwad: @upadhyeammit - No plan to add a function as iirc at the moment no update call on account table fields is available in ros-backend. Please correct if it exists anywhere.

upadhyeammit: I have few thoughts,
1. If updated_at is not being updated by any of the actions from the account/user then do we need this field? As then its copy of created_at?
2. If the intention is to see if account/user is involved in using the service considering the fact that we receive the update in the form of pcp data or host data, then maybe updated_at should be updated on every new upload of data from same account? however it will be too many of update calls? then maybe we can just see when was the last report uploaded by any of the systems from customer? and that should be fine?

kgaikwad: @upadhyeammit  - Removed `updated_at` field for now as no update call in ros-backend.

upadhyeammit (approved):

migrations/versions/52381cafc36f_add_created_at_field_to_rh_accounts.py
```diff
@@ -0,0 +1,32 @@
...
+revision = '52381cafc36f'
+down_revision = '020def9f5b94'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('rh_accounts', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('created_at', sa.DateTime(), nullable=True))
```
upadhyeammit, line 22: A thought after approval, do we need index for this column? if we are going to pull the data from table for metrics creation based on different time and date slices then I think its beneficial?
Maybe BRIN type index? https://www.postgresql.org/docs/current/brin-intro.html

kgaikwad: +1 Makes sense. I have added index of type `brin` on `created_at` column. Please request to review and merge if no additional change needed. Thanks!

PR #479, opened 2024-09-27
RHINENG-12954 retrieve operating_system using get method

r14chandra:


ros/processor/inventory_events_consumer.py
```diff
@@ -156,7 +156,7 @@ def process_system_details(self, msg):
                     "fqdn": host['fqdn'],
                     "cloud_provider": host['system_profile']['cloud_provider'],
                     "stale_timestamp": host['stale_timestamp'],
-                    "operating_system": host['system_profile']['operating_system'],
+                    "operating_system": host['system_profile'].get('operating_system'),
```
kgaikwad, line 159: In what situations, host will not have `operating_system` value?

kgaikwad: Hello @r14chandra  - I observed that in case of Inventory message of event type `created` ,  ROS do check for valid `operating_system` value here - https://github.com/RedHatInsights/ros-backend/blob/main/ros/lib/utils.py#L109
whereas for `updated` event type, it looks like ROS receiving payload that won't include `operating_system` value.
Instead of setting its value to `None`, can you ONLY assign `operating_system` value when present under `system_profile`. does it make sense to you? Thoughts?

kgaikwad: >  However, we are not dependent on this field for generating recommendations. 

As mentioned ROS RHEL is having validation check for this field. Could you please check on it. 

Specifically for updated event type, when ROS is having a system with OS info already in DB,  at least I don't see a point to override the value of the OS field when the key itself is not present in message.

PR #480, opened 2024-10-03
RHINENG-12954 Don't override OS and cloud provider value

r14chandra:


upadhyeammit (approved):

ros/processor/inventory_events_consumer.py
```diff
@@ -149,16 +149,22 @@ def process_system_details(self, msg):
         host = msg['host']
         with app.app_context():
             # 'platform_metadata' field not included when the host is updated via the API.
-            if (msg.get('type') == 'updated' and is_platform_metadata_check_pass(msg)):
```
kgaikwad, line 152: @r14chandra , what's the reason behind removing `is_platform_metadata_check_pass(msg)` check from here? Was it relevant to original change?

r14chandra: Redundant check. [system_allowed_in_ros](https://github.com/r14chandra/ros-backend/blob/5869facf1a77cc1f8c98791ee5e52d812b053495/ros/lib/utils.py#L361-L364) function already has this check.

kgaikwad: ahh okies. @r14chandra  can you please remove comment from above as it also talks related to `platform_metadata`. Thanks!

PR #507, opened 2024-11-28
RHINENG-14556 remove old code to keep the new service independent

r14chandra:

Eliminating codes that are not necessary for the new backend.

- [ ] Yes
- [X] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [X] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:


pr_check.sh
```diff
@@ -25,4 +25,11 @@ source $CICD_ROOT/build.sh
 source $CICD_ROOT/deploy_ephemeral_env.sh
 
 # Run iqe-ros somke tests with ClowdJobInvocation
-source $CICD_ROOT/cji_smoke_test.sh
+# source $CICD_ROOT/cji_smoke_test.sh  # require revisit after the new backed is up
```
upadhyeammit, line 28:
```suggestion
# source $CICD_ROOT/cji_smoke_test.sh  # require revisit after the new backend is up
```

upadhyeammit:
@r14chandra I can see we have not touched the models, so with this pr my impression is we go with current models as is and later on tweak it as per requirements?

r14chandra:
> @r14chandra I can see we have not touched the models, so with this pr my impression is we go with current models as is and later on tweak it as per requirements?

Yes, that's true. It’s also possible that what I am removing right now may be needed later.

upadhyeammit (approved):

PR #528, opened 2024-12-10
RHINENG-14812 remove database use in clowdapp

r14chandra:


r14chandra:
These resources are still being called in some places of the code. Need to close.

PR #559, opened 2024-12-31
RHINENG-13972 download and extract archive

r14chandra:

Listens to the Inventory topic (`platform.inventory.events`) and downloads archives based on `is_ros_v2` and `is_pcp_raw_data_collected` flags.

- [ ] Yes
- [X] No

## Security Checklist :lock:

- [ ] Bugfix
- [X] New Feature
- [ ] Refactor
- [X] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/ros/processor/suggestions_engine.py
+++ b/ros/processor/suggestions_engine.py
@@ -1,7 +1,11 @@
-import time
+import json
+import requests
+from http import HTTPStatus
+from ros.lib import consume
 from ros.lib.config import get_logger
+from tempfile import NamedTemporaryFile
 from prometheus_client import start_http_server
-from ros.lib.config import ROS_SUGGESTIONS_ENGINE_PORT
+from ros.lib.config import INVENTORY_EVENTS_TOPIC, METRICS_PORT
 
 
 logging = get_logger(__name__)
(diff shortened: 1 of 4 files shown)


ros/processor/suggestions_engine.py
```diff
@@ -1,26 +1,93 @@
...
+                logging.info(
+                    f"{self.service} - {self.event} - Downloaded the report successfully for system {host.get('id')}"
+                )
+                tempfile.flush()
+
+    def handle_delete(self, payload):
         pass
 
     def run(self):
+        logging.info(f"{self.service} - Processor is running. Awaiting msgs.")
```
upadhyeammit, line 56: Maybe `Engine is running`?

ros/processor/suggestions_engine.py
```diff
@@ -1,26 +1,93 @@
...
+                message = self.consumer.poll(timeout=1.0)
+                if message is None:
+                    continue
+
+                try:
+                    payload = json.loads(message.value().decode('utf-8'))
+                    event_type = payload['type']
+
+                    if 'delete' == event_type:
+                        self.handle_delete(payload)
```
upadhyeammit, line 68: We won't handle the delete events in this service?

ros/processor/suggestions_engine.py
```diff
@@ -1,26 +1,93 @@
...
+        response = requests.get(archiveURL, timeout=10)
+
+        if response.status_code != HTTPStatus.OK:
+            logging.error(
+                f"{self.service} - {self.event} - Unable to download the report for system {host.get('id')}. "
+                f"ERROR - {response.reason}"
+            )
+        else:
+            with NamedTemporaryFile(delete=True) as tempfile:
+                tempfile.write(response.content)
```
upadhyeammit, line 46: This is just downloading file and writing it to temp file, are we going to have separate pull request for extracting the archive?

Is it possible to separate out logic about downloading, extracting the archive?

upadhyeammit (changes requested):

ros/processor/suggestions_engine.py
```diff
@@ -1,26 +1,97 @@
...
+                    payload = json.loads(message.value().decode('utf-8'))
+                    event_type = payload['type']
+
+                    if 'created' == event_type or 'updated' == event_type:
+                        self.handle_create_update(payload)
+
+                except json.JSONDecodeError as error:
+                    logging.error(f"{self.service} - {self.event} - Failed to decode message: {error}")
+                except Exception as error:
+                    logging.error(f"{self.service} - {self.event} - Error processing message: {error}")
```
upadhyeammit, line 59: We are using the manual commit and so its required to commit the message after this try block?
It seems there are possibilities of we are able to read the message, however in later stages we fail somewhere,

In all of these situations it looks like at least we should have some well defined retry logic,

1. download
2. extract
3. running pmlogextract
4. running pmlogsummary
5. running the rules
6. sending message to kafka

I like this article about [timeouts](https://aws.amazon.com/builders-library/timeouts-retries-and-backoff-with-jitter/)

The timeout won't solve the whole situation and mostly we need something like,

1. A separate queue, maybe failed queue where we will store the message if we see failures in subsequent steps, and we can approach those messages with dedicated service.
2. If we don't want to go with separate queue, then all of above states will need finite set of exceptions in which we won't commit the kafka message. This brings another problem though, and that is, the message is valid, however its our services which are failing, that would mean, there is possibility of other messages getting processed without problem. In this approach we won't consume the next message, and it will halt the processing altogether.

I am not expecting to handle above two situation with work we are doing, however I do think we should have separate Jira card and mostly another RFC about with what strategy we are going to deal this situation.

P.S we should at least have the good retry logic!

r14chandra: The proposed approach would require additional setup and resources, including the creation of a new Kafka topic. It also requires monitoring and would add complexity to the implementation, which I believe adds unnecessary overhead. And we might still end up doing the same thing with this approach i.e. reprocessing the failed messages.

I suggest implementing a limited retry mechanism. For instance, we could allow up to three retries to avoid infinite loops before discarding the message. I think this limited retry approach is an efficient and straightforward solution that does not require any additional resources. I had implemented this for Puptoo and Engine also follows the same.

r14chandra: But I still don't understand its need for our application. Is it really required? Is it mandated not to lose data for our application? I think this approach should be used for critical applications.

As I mentioned earlier, we would still reprocess the failed messages (with the same logic?). However, if a message is corrupt or invalid, it could lead to continuous retry, causing an infinite loop.  

I feel that this would only introduce unnecessary complexity to our application. I think our approach should depend on the criticality of message processing.

upadhyeammit: I think yes we should have retry logic as well as separate queue in future to handle situations where message was not corrupt, however we failed to process it to the completion. This is because if any of the logical components start failing we just end up processing thousands of messages however we will miss to produce viable data from those?

upadhyeammit (approved):

ros/processor/suggestions_engine.py
kgaikwad: Can we have a JIRA ticket to address this delete event handling if we are skipping it for now? Thanks!

PR #576, opened 2025-01-21
RHINENG-13973 execute pcp commands

r14chandra:

Run `pmlogextract` on the extracted `.index` file, followed by `pmlogsummary` on the resulting output file.

- [ ] Yes
- [X] No

## Security Checklist :lock:

- [ ] Bugfix
- [X] New Feature
- [X] Refactor
- [X] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/ros/lib/pcp_extract_config
+++ b/ros/lib/pcp_extract_config
@@ -1,0 +1,10 @@
+disk.dev.total
+hinv.ncpu
+kernel.all.cpu.idle
+kernel.all.pressure.cpu.some.avg
+kernel.all.pressure.io.full.avg
+kernel.all.pressure.io.some.avg
+kernel.all.pressure.memory.full.avg
+kernel.all.pressure.memory.some.avg
+mem.physmem
+mem.util.available
--- a/ros/processor/suggestions_engine.py
+++ b/ros/processor/suggestions_engine.py
@@ -1,15 +1,21 @@
+import os
 import json
-import requests
+import shutil
+import subprocess
 from http import HTTPStatus
-from ros.lib import consume
-from insights import extract
-from ros.lib.config import get_logger
+from contextlib import contextmanager
 from tempfile import NamedTemporaryFile
+
+import requests
+from insights import extract
 from prometheus_client import start_http_server
+
+from ros.lib import consume
 from ros.lib.config import (
-    INVENTORY_EVENTS_TOPIC,
+    get_logger,
     METRICS_PORT,
-    GROUP_ID_SUGGESTIONS_ENGINE
+    INVENTORY_EVENTS_TOPIC,
+    GROUP_ID_SUGGESTIONS_ENGINE,
 )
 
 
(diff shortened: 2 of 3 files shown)


ros/processor/suggestions_engine.py
```diff
@@ -1,20 +1,39 @@
...
     METRICS_PORT,
-    GROUP_ID_SUGGESTIONS_ENGINE
+    INVENTORY_EVENTS_TOPIC,
+    GROUP_ID_SUGGESTIONS_ENGINE,
 )
 
 
 logging = get_logger(__name__)
 
+PCP_METRICS = [
```
upadhyeammit, line 24: This should not be required as we are extracting the metrics only those metrics will be present in extracted file,
~~~
$ pmlogextract -c pcp-extract-config /var/log/pcp/pmlogger/fedora/20250127.index /tmp/pmlogextract/

$ pmlogsummary -f .index 
disk.dev.total	["nvme0n1"]	44.908	count / sec
kernel.all.cpu.idle		15.395	none
kernel.all.pressure.cpu.some.avg	["10 second"]	0.496	none
kernel.all.pressure.cpu.some.avg	["1 minute"]	0.502	none
kernel.all.pressure.cpu.some.avg	["5 minute"]	0.488	none
kernel.all.pressure.memory.full.avg	["10 second"]	0.000	none
kernel.all.pressure.memory.full.avg	["1 minute"]	0.000	none
kernel.all.pressure.memory.full.avg	["5 minute"]	0.000	none
kernel.all.pressure.io.some.avg	["10 second"]	0.007	none
kernel.all.pressure.io.some.avg	["1 minute"]	0.004	none
kernel.all.pressure.io.some.avg	["5 minute"]	0.002	none
kernel.all.pressure.io.full.avg	["10 second"]	0.006	none
kernel.all.pressure.io.full.avg	["1 minute"]	0.004	none
kernel.all.pressure.io.full.avg	["5 minute"]	0.001	none
hinv.ncpu		16.000	none
mem.util.available		56882700.434	Kbyte
~~~

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
+            f"{service} - {event} - insights_archive.txt not found in the extracted dir for system {host.get('id')}."
+        )
+        return None
+
+    pmlogger_dir = os.path.join(extracted_dir_root, "data/var/log/pcp/pmlogger/")
+
+    index_files = [
+        file for file in os.listdir(pmlogger_dir) if file.endswith(".index")
+    ]
+    index_file_path = os.path.join(pmlogger_dir, index_files[0])
```
upadhyeammit, line 121: Maybe we can avoid this if we create named files when executing the `pmlogextract`,
~~~
$ pmlogextract -c pcp-extract-config /var/log/pcp/pmlogger/fedora/20250127.index /tmp/pmlogextract/20250127
$ ls /tmp/pmlogextract/20250127.*
/tmp/pmlogextract/20250127.0  /tmp/pmlogextract/20250127.index  /tmp/pmlogextract/20250127.meta
~~~

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
+
+    return index_file_path
+
 
-    response = requests.get(archive_URL, timeout=10)
+def create_output_dir(request_id, host):
+    output_dir = f"/tmp/pmlogextract-output-{request_id}/"
```
upadhyeammit, line 127: Though we need the directory and files for temporary purpose however these files are getting passed to next command `pmlogsummary`, so in this situation I don't think using `/tmp` is wise idea. The initial impression says the `/tmp` is managed by `systemd-tmpfiles-clean.service` so its possible that these could be deleted before `pmlogsummary` requires it.
Should we look into adding small persistent volume? maybe 250Mb is enough? and there we do these file operations?

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
+            logging.info(f"{service} - {event} - Successfully ran pmlogextract command for system {host.get('id')}.")
+        except subprocess.CalledProcessError as error:
+            logging.error(
+                f"{service} - {event} - Error running pmlogextract command for system {host.get('id')}: {error.stdout}"
+            )
+            return
+
+        pmlogsummary_command = [
+            "pmlogsummary",
+            "-f",
```
upadhyeammit, line 161: Do you think maybe -F(CSV) format can make our life easy when parsing the output,
~~~
$ pmlogsummary -F 20250127.index
disk.dev.total,["nvme0n1"],76.318,count / sec
kernel.all.cpu.idle,,11.194,none
kernel.all.pressure.cpu.some.avg,["10 second"],0.533,none
kernel.all.pressure.cpu.some.avg,["1 minute"],0.619,none
kernel.all.pressure.cpu.some.avg,["5 minute"],0.571,none
kernel.all.pressure.memory.full.avg,["10 second"],0.000,none
kernel.all.pressure.memory.full.avg,["1 minute"],0.000,none
kernel.all.pressure.memory.full.avg,["5 minute"],0.000,none
kernel.all.pressure.io.some.avg,["10 second"],0.019,none
kernel.all.pressure.io.some.avg,["1 minute"],0.019,none
kernel.all.pressure.io.some.avg,["5 minute"],0.013,none
kernel.all.pressure.io.full.avg,["10 second"],0.019,none
kernel.all.pressure.io.full.avg,["1 minute"],0.018,none
kernel.all.pressure.io.full.avg,["5 minute"],0.012,none
hinv.ncpu,,16.000,none
mem.util.available,,56255058.797,Kbyte
mem.physmem,,64422276.000,Kbyte
~~~

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
+            os.path.join(output_dir, ".index"),
+            *PCP_METRICS
+        ]
+
+        logging.info(f"{service} - {event} - Running pmlogsummary command for system {host.get('id')}.")
+        try:
+            subprocess.run(pmlogsummary_command, check=True, capture_output=True, text=True)
```
upadhyeammit, line 168: Should we save the output to some variable to use it? or you think this should be done in another pr?
like,
~~~
result = subprocess.run(pmlogsummary_command, check=True, capture_output=True, text=True)
print(result.stdout)
print(result.stderr)
~~~

ros/lib/pcp_extract_config
```diff
@@ -0,0 +1,10 @@
+disk.dev.total
+hinv.ncpu
+kernel.all.cpu.idle
+kernel.all.pressure.cpu.some.avg
+kernel.all.pressure.io.full.avg
+kernel.all.pressure.io.some.avg
+kernel.all.pressure.memory.full.avg
+kernel.all.pressure.memory.some.avg
+mem.physmem
+mem.util.available
```
upadhyeammit, line 10: This file is very specific to a processor, is it possible to create helper directory for the processor itself and then save it there?

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
+
+        logging.info(f"{service} - {event} - Running pmlogextract command for system {host.get('id')}.")
+        try:
+            subprocess.run(pmlogextract_command, check=True)
+            logging.info(f"{service} - {event} - Successfully ran pmlogextract command for system {host.get('id')}.")
+        except subprocess.CalledProcessError as error:
+            logging.error(
+                f"{service} - {event} - Error running pmlogextract command for system {host.get('id')}: {error.stdout}"
+            )
+            return
```
upadhyeammit, line 157: Just thought we can have separate methods for `pmlogextract` and `pmlogsummary`?

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
-    else:
-        with NamedTemporaryFile(delete=True) as tempfile:
-            tempfile.write(response.content)
-            logging.info(
-                f"{service} - {event} - Downloaded the report successfully for system {host.get('id')}"
+    finally:
+        try:
+            if os.path.exists(output_dir):
+                shutil.rmtree(output_dir)
+                logging.info(f"{service} - {event} - Cleaned up output directory for system {host.get('id')}.")
```
upadhyeammit, line 184: Do we need this as `info` message?

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
+        except Exception as error:
+            logging.error(
+                f"{service} - {event} - Error cleaning up the output directory for system {host.get('id')}: {error}"
```
upadhyeammit, line 187: same, maybe not required as `info` message?

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
+            return
+
+        pmlogsummary_command = [
+            "pmlogsummary",
+            "-f",
+            os.path.join(output_dir, ".index"),
+            *PCP_METRICS
+        ]
+
+        logging.info(f"{service} - {event} - Running pmlogsummary command for system {host.get('id')}.")
```
upadhyeammit, line 166: Maybe not required as `info` message

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
+
+        pmlogextract_command = [
+            "pmlogextract",
+            "-c",
+            "ros/lib/pcp_extract_config",
+            index_file_path,
+            output_dir
+        ]
+
+        logging.info(f"{service} - {event} - Running pmlogextract command for system {host.get('id')}.")
```
upadhyeammit, line 149: Maybe not required as info message

ros/processor/suggestions_engine.py
```diff
@@ -64,25 +93,125 @@ def run(self):
...
+                logging.info(f"{service} - {event} - Cleaned up output directory for system {host.get('id')}.")
+        except Exception as error:
+            logging.error(
+                f"{service} - {event} - Error cleaning up the output directory for system {host.get('id')}: {error}"
+            )
+
+
+@contextmanager
+def download_and_extract(service, event, archive_URL, host, org_id):
+    logging.info(f"{service} - {event} - Downloading the report for system {host.get('id')}.")
```
upadhyeammit, line 193: Maybe not required as info message

upadhyeammit (changes requested):

ros/processor/suggestions_engine.py
r14chandra: I think you overlooked this part. Here, we want to get the `index` file from the extracted archive itself. We are not responsible for naming those files.

ros/processor/suggestions_engine.py
r14chandra: In that case, we can use `/var/tmp` .

ros/lib/pcp_extract_config
r14chandra: That would still be under the lib directory where there are files that are being used by a specific processor like `aws_instance_types.py`. 
Once we are done with the development, we may consider keeping it under the processor directory itself along with other related files. We can skip it for now.

ros/processor/suggestions_engine.py
upadhyeammit: I need to check! If I understand correctly the insights-core does not have duties about collecting data from containers? those are instances of VM?

upadhyeammit: The `/var/tmp` is better choice, I do see security concerns as well when using these directories,
https://www.redhat.com/en/blog/polyinstantiating-tmp-and-vartmp-directories

I have captured both the usage and the security concern in new Jira ticket RHINENG-15674 and we can evaluate it later. As of now lets go with `/var/tmp`.

ros/processor/suggestions_engine.py
```diff
@@ -22,6 +28,141 @@ def __init__(self):
...
+
+    def run_pcp_commands(self, host, index_file_path, request_id):
+        sanitized_request_id = request_id.replace("/", "_")
+        output_dir = self.create_output_dir(sanitized_request_id, host)
+
+        self.run_pmlogextract(host, index_file_path, output_dir)
+
+        self.run_pmlogsummary(host, output_dir)
+
+        logging.info(f"{self.service} - {self.event} - Successfully ran pcp commands for system {host.get('id')}.")
```
upadhyeammit, line 92: I think your previous approach was correct, as I was in doubt I ended up doing this locally and it shows, if `run_pmlogextract` fails still we end up processing the `run_pmlogsummary`. Though `run_pmlogsummary` will throw exception as well however we can avoid that. So can we chain the exceptions again?

upadhyeammit (approved):

upadhyeammit (approved):

PR #593, opened 2025-02-24
Testing konflux

akhil-jha:


akhil-jha:
The deployment requires gitlab access. Needs RHCLOUD-35842 similar to other services.

PR #625, opened 2025-04-29
RHINENG-15101 Handle scenario where is_ros_v2 true but no raw data

r14chandra:

Resolves [RHINENG-15101](https://issues.redhat.com/browse/RHINENG-15101)

- [ ] Yes
- [X] No

## Security Checklist :lock:

- [ ] Bugfix
- [X] New Feature
- [ ] Refactor
- [X] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:


scripts/no_pcp_message.json
upadhyeammit, line 1: This file can belong to `sample-files`? The filename can be improved something like, `no_pcp_raw_data_collected`? considering `pcp` is name of the package as well.

ros/processor/suggestions_engine.py
```diff
@@ -174,23 +176,32 @@ def download_and_extract(self, archive_URL, host, org_id):
...
-        if not self.is_pcp_collected(platform_metadata):
+        if not is_ros_enabled:
+            return
+
+        if not is_pcp_collected:
+            logging.debug(
+                f"{self.service} - {self.event} - Triggering an event for system {host.get('id')}"
+            )
+            self.consumer.commit()
+            produce_report_processor_event(payload, platform_metadata, self.producer)
```
upadhyeammit, line 204: The `payload` does contain the `platform_metadata`? 
We do have data contract, so in that context should we only send the fields which make sense to consume by our db schema? This will reduce the size of the message and avoid sending data which is not useful?

upadhyeammit (changes requested):

ros/processor/suggestions_engine.py
r14chandra: That means we need to create a specific payload for all the cases we handle where `is_ros_v2` is `true` and `is_pcp_raw_collected` can be `true` or `false`, and events from the API. 
While not complicating the Kafka message schema, the data contract should be considered while working with the fields to be saved in the DB, specifically under report_processor.

PR #832, opened 2026-04-27
RHINENG-25817 use ssl mode from clowder

r14chandra:

As a part of [RHINENG-25817](https://redhat.atlassian.net/browse/RHINENG-25817)

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [X] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## ROS RHEL GitHub label usage for executing a desired test suite

Select the appropriate GitHub label to control which ROS RHEL backend tests run for this PR (Labels can be added before or after commits are pushed):

**Note:** Changing a Label on a PR with an already-tested commit will not trigger a new test run

**Available ROS RHEL labels:**

- **test-backend-v1:** Run legacy backend tests only
- **test-backend-v2:** Run new backend tests only
- **test-backend-both:** Run both backend tests

## Additional :mega:

[RHINENG-25817]: https://redhat.atlassian.net/browse/RHINENG-25817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Enhancements:
- Derive database SSL mode from Clowder configuration when available and append the appropriate sslmode parameter to the database URI when no certificate path is configured.

### DIFF
--- a/ros/lib/config.py
+++ b/ros/lib/config.py
@@ -129,11 +129,10 @@ else:
     UNLEASH_URL = os.getenv("UNLEASH_URL", "http://localhost:3063/api")
 
 DB_URI = f"postgresql://{DB_USER}:{DB_PASSWORD}"\
-         f"@{DB_HOST}:{DB_PORT}/{DB_NAME}"
-if DB_SSL_CERTPATH:
-    DB_URI += f"?sslmode={DB_SSL_MODE}&sslrootcert={DB_SSL_CERTPATH}"
-elif DB_SSL_MODE != "disable":
-    DB_URI += f"?sslmode={DB_SSL_MODE}"
+         f"@{DB_HOST}:{DB_PORT}/{DB_NAME}"\
+         f"?sslmode={DB_SSL_MODE}"
+if DB_SSL_MODE != "disable" and DB_SSL_CERTPATH:
+    DB_URI += f"&sslrootcert={DB_SSL_CERTPATH}"
 
 DB_POOL_SIZE = int(os.getenv("DB_POOL_SIZE", '5'))
 DB_MAX_OVERFLOW = int(os.getenv("DB_MAX_OVERFLOW", '10'))
--- a/ros/lib/config.py
+++ b/ros/lib/config.py
@@ -31,7 +31,7 @@ def build_endpoint_url(ep):
 
 LOG = logging.getLogger(__name__)
 CLOWDER_ENABLED = True if os.getenv("CLOWDER_ENABLED", default="False").lower() in ["true", "t", "yes", "y"] else False
-DB_SSL_MODE = "verify-full"
+DB_SSL_MODE = "disable"
 DB_SSL_CERTPATH = None
 
 if CLOWDER_ENABLED:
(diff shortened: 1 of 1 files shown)


ros/lib/config.py
```diff
@@ -45,6 +45,7 @@ def build_endpoint_url(ep):
     DB_PASSWORD = LoadedConfig.database.password
     DB_HOST = LoadedConfig.database.hostname
     DB_PORT = LoadedConfig.database.port
+    DB_SSL_MODE = LoadedConfig.database.sslMode
```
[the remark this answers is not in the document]
r14chandra: This is consistent with how all other Clowder config values are handled in this file

ros/lib/config.py
```diff
@@ -131,6 +132,8 @@ def build_endpoint_url(ep):
          f"@{DB_HOST}:{DB_PORT}/{DB_NAME}"
 if DB_SSL_CERTPATH:
     DB_URI += f"?sslmode={DB_SSL_MODE}&sslrootcert={DB_SSL_CERTPATH}"
```
kgaikwad: @r14chandra - I noticed `sslmode` parameter is being used in this earlier check. Should we consolidate or refactor these conditions to avoid duplication or conflicting logic?

ros/lib/config.py
```diff
@@ -128,9 +129,10 @@ def build_endpoint_url(ep):
     UNLEASH_URL = os.getenv("UNLEASH_URL", "http://localhost:3063/api")
 
 DB_URI = f"postgresql://{DB_USER}:{DB_PASSWORD}"\
-         f"@{DB_HOST}:{DB_PORT}/{DB_NAME}"
-if DB_SSL_CERTPATH:
-    DB_URI += f"?sslmode={DB_SSL_MODE}&sslrootcert={DB_SSL_CERTPATH}"
+         f"@{DB_HOST}:{DB_PORT}/{DB_NAME}"\
+         f"?sslmode={DB_SSL_MODE}"
```
kgaikwad: Based on current behavior, it doesn't seem to be required query param. Any specific reason you are changing current behavior?

ros/lib/config.py
kgaikwad: @r14chandra - Instead of clubbing these two query parameters, can you handle those separately based on presence and constant value?

ros/lib/config.py
```diff
@@ -129,8 +130,10 @@ def build_endpoint_url(ep):
 
 DB_URI = f"postgresql://{DB_USER}:{DB_PASSWORD}"\
          f"@{DB_HOST}:{DB_PORT}/{DB_NAME}"
-if DB_SSL_CERTPATH:
-    DB_URI += f"?sslmode={DB_SSL_MODE}&sslrootcert={DB_SSL_CERTPATH}"
+if DB_SSL_MODE != "disable":
+    DB_URI += f"?sslmode={DB_SSL_MODE}"
+    if DB_SSL_CERTPATH:
+        DB_URI += f"&sslrootcert={DB_SSL_CERTPATH}"
```
kgaikwad, line 138: Based on AWS document - https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL.Concepts.General.SSL.html, the client requires CA certificates only when sslmode is set to verify-ca or verify-full.
Could you please update the condition accordingly so that sslrootcert is included only for these modes?

```suggestion

if DB_SSL_MODE != "disable":
    DB_URI += f"?sslmode={DB_SSL_MODE}"
# Only attach sslrootcert when verification is required 
if DB_SSL_MODE in ("verify-ca", "verify-full"): 
     if not DB_SSL_CERTPATH: 
              raise ValueError( "sslrootcert is required when sslmode is verify-ca or verify-full" ) 
     DB_URI += f"&sslrootcert={DB_SSL_CERTPATH}"
```

kgaikwad (approved):

